PR #175, opened 2026-02-01
Prepare 0.3.4 with visu adjustments

Apollon77:


packages/dashboard/src/pages/components/node-details.ts
```diff
@@ -79,7 +79,7 @@ export class NodeDetails extends LitElement {
             <md-list>
                 <md-list-item>
                     <div slot="headline">
-                        <b>Node ${this.node.node_id} ${this.node.nodeLabel}</b>
+                        <b>${this.node.nodeLabel || "Node Info"}</b>
```
ximex: use `??` instead of `||`

PR #215, opened 2026-02-07
Add childLock attribute to EveCluster

lboue:
Add childLock attribute to EveCluster

Feature request: [Eve Energy: Add support for Child Lock control in Matter integration #1099 ](https://github.com/orgs/home-assistant/discussions/1099)

Prerequisites:
- https://github.com/matter-js/matter.js/pull/3192


lboue:
> Thanks,
> 
> Is this a writable attribute? Then we need more things to happen before... Additionally adding it here will not enable the python client in HA to also know it currently!

Yes, that's correct.

PR #219, opened 2026-02-08
Actively listen for mDNS announcements for devices in WaitingForDeviceDiscovery

c10l:
In some circumstances, the mDNS scan for devices in `WaitingForDeviceDiscovery` may not be triggered, which can lead to devices not being discovered or taking a long time to be discovered.

This can happen, for example, when the device is already in the cache and the cache is not expired, or when the device is not responding to mDNS queries.

This commit adds an active mDNS scan to ensure that devices in `WaitingForDeviceDiscovery` are properly discovered.

The second commit improves response times to mDNS announcements by reducing timeouts and parallelising device discoveries.

FULL DISCLAIMER: This code was implemented by Claude Code, with my supervision. I have read the entirety of the change, but I am not a JavaScript or TypeScript developer, and this is my first time interacting with this codebase.

### DIFF
--- a/packages/ws-controller/src/controller/MdnsDeviceScanner.ts
+++ b/packages/ws-controller/src/controller/MdnsDeviceScanner.ts
@@ -35,6 +35,8 @@ export interface MdnsDeviceScannerDeps {
     getNode(nodeId: NodeId): PairedNode;
     /** Discover a node's operational addresses via mDNS with a short timeout. */
     findDevice(nodeId: NodeId): Promise<DiscoveredDevice | undefined>;
+    /** Delay in ms between scanning individual nodes. Defaults to 1000. */
+    interNodeDelayMs?: number;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


c10l:
Hi @Apollon77 ! Thanks for reviewing. The issue was originally reported [on Discord](https://discord.com/channels/330944238910963714/1461119138713047274/1470122177536725064). It happened for a few devices, but I chased on specifically that's on the screenshots.

Basically the Matter Server was stuck, likely trying to connect (unfortunately even with debug logs it only seems to log once it times out, not before the attempt, so it's difficult to ascertain). It was many minutes after the device had made an mDNS announcement that the Matter Server finished timing out for the old IPs and attempted the new one.

This change would force the Matter Server to drop the wait and attempt a connection on the new IP immediately (or within seconds) of a new mDNS announcement for a known device.

If it should be implemented elsewhere, I'm also happy to take a stab at it.

PR #220, opened 2026-02-08
Fix duplicate entities caused by cache replacement

c10l:
The root cause is in `AttributeDataCache.#populateFromNode()` (line 130 of `AttributeDataCache.ts`): it creates a **brand new `Map()`** and replaces the entire cache entry, discarding any previously-known endpoints/attributes that aren't present in the current endpoint tree.

When a `structureChanged` or `stateChanged(Connected)` event fires while a device's endpoint tree is incomplete (e.g., the switch endpoint is ready but the energy meter clusters are still initializing), the cache replacement drops the energy meter attributes. The subsequent `node_updated` event sent to HA has no energy meter data, causing HA to mark those entities as unavailable. When the endpoints later appear and the cache is rebuilt with the full set, HA creates duplicate entities with `_2` suffixes.

DISCLAIMER: The code was written by Claude. I have presented it with the failure conditions, it wrote the tests and showed them failing, proving the hypothesis, then wrote the code to fix it. I have reviewed the code and tests but I am not proficient in JavaScript or TypeScript.

image

### DIFF
--- a/packages/ws-controller/src/controller/AttributeDataCache.ts
+++ b/packages/ws-controller/src/controller/AttributeDataCache.ts
@@ -125,7 +125,9 @@ export class AttributeDataCache {
 
     /**
      * Populate the cache for a node from its current state.
-     * Creates a completely fresh cache structure.
+     * Merges freshly-collected endpoint data into the existing cache so that
+     * previously-known endpoints are preserved when they are temporarily
+     * unavailable (e.g., during reconnection when not all endpoints are ready).
      */
     #populateFromNode(node: PairedNode): void {
         const nodeId = node.nodeId;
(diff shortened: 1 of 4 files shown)


Apollon77:
Please upload the full log because I can not see the fill picture (please never grep logs if devs do not ask for it because honestly you do not know what's important to filter). For me it seems that somethings happened 12:17 because it seems that the whole server did restart. 

Basically only the electrical values changed (12:17:37.709) on subscribe ...

But log is not containing the needed details to know when what was sent to the python server ...

Then 12:19:05.908 another restart of the whole server, again correctly resubscribed, no electrical values changed

So from the partial log all looks fine and "as expected". maybe there is an issue if the server crashes? But I guess you want to look into that? Maybe a RAM issue or such?

PR #412, opened 2026-03-22
Feature: allow network interface name for Websocket binding

FuNK3Y:
Allows to specify a network interface name enclosed in `{{ }}` as the listen address. The network interface name will be substituted with the primary ipv4 address of that given interface.

This is particularly useful when the IP you want to listen to is dynamic. It can typically happen if your host is multihomed and that you do not want to listen to every interfaces. Or if you want to use macvlan networking for line with matter devices of sight while binding the socket to a bridged interface.

This would go in the sense of https://github.com/home-assistant/architecture/discussions/1034

### DIFF
--- a/packages/matter-server/src/cli.ts
+++ b/packages/matter-server/src/cli.ts
@@ -226,9 +226,7 @@ export function parseCliArgs(argv?: string[]): CliOptions {
                 );
 
                 if (normalizedInterfaceAddresses.length === 0) {
-                    throw new InvalidArgumentError(
-                        `No valid IP address found for interface ${interfaceName}`,
-                    );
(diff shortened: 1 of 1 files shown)


packages/matter-server/src/cli.ts
```diff
@@ -207,6 +207,25 @@ export function parseCliArgs(argv?: string[]): CliOptions {
         listenAddress = [process.env.LISTEN_ADDRESS];
     }
 
+    // Substitute {{interface}} patterns with its primary IP addresse
```
Apollon77:
```suggestion
    // Substitute {{interface}} patterns with its primary IP address
```

FuNK3Y:
Done @Apollon77 !

PR #439, opened 2026-03-24
Fix node_removed event sent before BasicInformation Leave event

### LINKED ISSUE #75: Change order for "node_removed" and Basic Information Cluster Leave events
### Version

0.2.6

### Node.js Version

N/A

### Operating System

Docker

### Issue Description

When a node is removed from the fabric from another controller like Apple Home these events are sent from matterjs-server to the client in this order:

1) "node_removed"

2) Basic Information Cluster Leave event

In order to guarantee cleanup of the matterjs-server clients datastructures for a node, the cleanup needs to be done in step 1 (since event 2 isn't triggered if the node is removed via websocket command while the device is offline). That means that it will be difficult to handle event 2) in the client, since the data structure for handling it was deleted when handling 1).

It would be more logical if the "node_removed" event was the last event sent  from matterjs-server related to this node as this would signal that everything in the matterjs-server has been cleaned up for that node and that no further messages is expected from that node.

### Steps to Reproduce

N/A

### Server Logs

```shell
N/A
```

### Device Information

N/A

### Additional Context

_No response_

### Pre-submission Checklist

markvp:
## Summary

- Defers `nodeDecommissioned` emission via `queueMicrotask()` so that events emitted synchronously in the same call stack (e.g. the BasicInformation Leave event that triggers decommissioning from another controller) are relayed to WebSocket clients **before** `node_removed`
- Cleans up internal state (`#nodes`, `#customClusterPoller`) in the `decommissioned` handler so that nodes removed by an external controller are cleaned up identically to the WebSocket `remove_node` command path
- Wraps the deferred microtask body in try/catch to prevent uncaught exceptions from crashing the process
- Adds unit tests for the decommission event ordering pattern

Fixes #75

## Details

When another controller (e.g. Apple Home) removes a node from the fabric, matter.js's `PairedNode` processes the Leave event in this order within a single call stack:

1. `_triggerEventUpdate()` fires the cluster-level Leave event → `Peers.#onLeave()` → `decommissioned.emit()` (synchronous)
2. `eventTriggered.emit()` fires the PairedNode-level event

Since `ControllerCommandHandler` relayed both synchronously, `node_removed` reached WebSocket clients before the Leave `node_event`. By deferring `nodeDecommissioned` to a microtask, the synchronous `eventTriggered` completes first.

### Internal state cleanup

Previously, the `decommissioned` event handler only emitted `nodeDecommissioned` to notify WebSocket clients but did not clean up `#nodes` or `#customClusterPoller`. This meant nodes removed by an external controller left stale entries in server state. The handler now mirrors the cleanup performed by `decommissionNode()`.

### Offline removal via WebSocket command

When a node is removed via the WebSocket `remove_node` command while the device is offline, no Leave event is emitted — only `node_removed`. This is correct behaviour because `events.decommissioned` is always emitted regardless of the removal path:

1. **External controller**: Leave event → `Peers.#onLeave()` → `decommissioned.emit()`
2. **WebSocket command (online)**: `PairedNode.decommission()` → `decommissioned.emit()`
3. **WebSocket command (offline)**: `PairedNode.close(true)` → `decommissioned.emit()`

The Leave event is purely informational for clients — the server performs no special processing on it. `node_removed` is the authoritative signal that cleanup is complete.

## Test plan

- [x] `npm run format` passes
- [x] `npm run lint` passes
- [x] `npm run build` passes
- [x] `npm test` — ws-controller (3/3 new + existing), ws-client (43/43) pass; integration test failures are pre-existing on main
- [x] New tests verify:
  - Event ordering: `node_event` (Leave) arrives before `node_removed` when both fire synchronously
  - Offline removal: `node_removed` is emitted even when no Leave event fires
  - Error resilience: listener exceptions are caught and don't crash the process

### DIFF
--- a/packages/ws-controller/src/controller/ControllerCommandHandler.ts
+++ b/packages/ws-controller/src/controller/ControllerCommandHandler.ts
@@ -361,7 +361,23 @@ export class ControllerCommandHandler {
             basicInfoChangedInBatch = false;
             this.events.nodeStructureChanged.emit(nodeId);
         });
-        node.events.decommissioned.on(() => this.events.nodeDecommissioned.emit(nodeId));
+        node.events.decommissioned.on(() => {
+            // Defer so that any events emitted synchronously in the same call stack
+            // (e.g. the BasicInformation Leave event that triggers decommissioning) are
+            // relayed to clients before node_removed.  See matter-js/matterjs-server#75.
+            queueMicrotask(() => {
+                try {
+                    // Clean up internal state — mirrors the cleanup in decommissionNode() so
+                    // that nodes removed by an external controller are handled identically.
+                    this.#nodes.delete(nodeId);
+                    this.#customClusterPoller.unregisterNode(nodeId);
+
+                    this.events.nodeDecommissioned.emit(nodeId);
+                } catch (error) {
+                    logger.error(`Error during decommission cleanup for node ${this.formatNode(nodeId)}`, error);
+                }
+            });
+        });
         node.events.nodeEndpointAdded.on(endpointId => this.events.nodeEndpointAdded.emit(nodeId, endpointId));
         node.events.nodeEndpointRemoved.on(endpointId => this.events.nodeEndpointRemoved.emit(nodeId, endpointId));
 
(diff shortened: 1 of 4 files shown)


Apollon77 (changes requested):
Honestly ... this PR is difficult because this test does not at all test the reality but just takes assumptions. So when we need to very this then via enhancing the integration test.

And even then I am unsure if I like this solution because this is still hacky and only works in very defined considerations that might not match the reality

PR #457, opened 2026-03-29
Improve CHIP SDK compatibility in Python cluster generator

TheJulianJES:
### Note: Target branch

This is a PR targeting the branch of the original PR:
- https://github.com/matter-js/matterjs-server/pull/303

This PR is to ease review of the differences compared to that PR and can later be merged into the `generated-python-client` branch when it's ready.

### TODOs

- ~~The comparison scripts included in this PR should be dropped before it's merged!~~
  - ~~Lint checks are expected to fail because of this~~
- Go through all "key fixes in generator" and see if we actually want them
  - E.g. we may want to include the `eventList` attribute
- Go through all "intentionally not changed" notes
- Go through all "Remaining known differences"
- ~~Test with Home Assistant Core~~ 
- ...?

## AI summary

Improves the Python cluster generator (`generate-python-clusters.ts`) to produce definitions that are much more closely aligned with the CHIP SDK's `chip.clusters.Objects`. A comparison script was used to systematically identify and fix differences across all 121 generated clusters. Fixes were verified by starting Home Assistant with the Matter integration.

### Key fixes in the generator

- **Fix list element optionality**: The CHIP SDK uses `List[Type]`, not `List[Type | None]`. List entry types are now resolved via `resolveScalarType()` directly, skipping optional/nullable wrappers (~183 annotations fixed).
- **Use `model.members` for command/event fields**: Commands and events now use `model.members` (which includes inherited fields) instead of `model.children`. This fixes empty `WithOnOff` commands in LevelControl and ~51 other missing command fields.
- **Fall back to base attribute when local override is an empty stub**: Derived clusters (e.g., DishwasherMode from ModeBase) can have local attribute stubs with no type or children that override the base cluster's full definition. Now falls back to the base attribute, fixing `supportedModes` resolving as `List[uint]` instead of `List[ModeOptionStruct]` across all mode clusters.
- **Deduplicate and sort attributes**: Attributes are now deduplicated by ID (fixes duplicate `minLevel` in LevelControl from the Matter.js model) and sorted by tag to match CHIP SDK ordering.
- **Fix multi-bit bitmap masks**: The generator only handled single-bit constraints (`{"value": 4}` → `1 << 4`). Multi-bit fields use range constraints (`{"min": 0, "max": 1}` → bits 0-1) which fell back to `0x1` instead of the correct mask. Example from `Thermostat.HVACSystemTypeBitmap`:
  - `CoolingStage` (bits 0-1): was `0x1`, now `0x3`
  - `HeatingStage` (bits 2-3): was `0x1`, now `0xC`
- **Use metatype as fallback when attribute type is undefined**: Attributes with no explicit type but a valid metatype (e.g., `float`, `boolean`, `string`) were all resolving to `uint`. Now checks metatype first. Fixes EveCluster `pressure`/`voltage`/`watt`/`altitude` typed as `uint` instead of `float32`, and BridgedDeviceBasicInformation `vendorName`/`reachable` etc. typed as `uint` instead of `str`/`bool`.
- **Use signed `int` for temperature, power, energy, voltage types**: Builds a `SIGNED_INT_TYPES` set by walking the Matter.js datatype hierarchy (global + cluster-level) — any type whose base resolves to `int8`/`int16`/.../`int64` is automatically recognized as signed. No hardcoded type aliases needed. Fixes Thermostat setpoints and calibration values, TemperatureMeasurement, DeviceEnergyManagement power values, and all other signed integer attributes that could have negative values.
- **Remove redundant quotes from type annotations**: The generated files use `from __future__ import annotations`, so explicit quoting (`value: 'uint' = 0`) is redundant. Now matches CHIP SDK style: `value: uint = 0`.
- **Remove incorrect `GroupId` field-name override**: CHIP SDK uses `groupID` (uppercase ID) in most command fields; the previous `GroupId: "groupId"` override was wrong for the majority of uses.
- **Fix FIELD_NAME_OVERRIDES keys to match Matter.js model names**: Override keys must use the raw model name, not the post-`toChipName` version. Fixes `requirePINforRemoteOperation` (DoorLock), `IPv4Addresses`/`offPremiseServicesReachableIPv4` (GeneralDiagnostics), `extendedPanIdPresent`/`panIdPresent` (ThreadNetworkDiagnostics).
- **Rename `objects/` to `cluster_defs/`**: The per-cluster files directory was renamed from `objects` to `cluster_defs` to avoid a macOS case-insensitivity collision (`objects/` and `Objects.py` resolved to the same path, causing pylint to read `__init__.py` instead of `Objects.py`).
- **Pass `clusterName` to all generators**: `toCamelCase` and `toChipClassName` were only receiving `clusterName` in some generators (cluster-level attributes, enums). Now all generators (structs, commands, events, bitmaps) pass it through, so cluster-qualified overrides in `FIELD_NAME_OVERRIDES` and `CLASS_NAME_OVERRIDES` work in all contexts.

### CHIP SDK naming compatibility

Added several override mechanisms to match CHIP SDK naming where it diverges from the Matter.js model:

- **`CLASS_NAME_OVERRIDES` + `toChipClassName()`**: New override map (supports both global and cluster-qualified `"Cluster.Name"` keys) and helper function for class names where `toChipName()` doesn't match the CHIP SDK. Applied to all class-name generation sites. Examples:
  - `CommissioningARL`, `AdminVendorId`, `ICACCSRRequest`, `ACCapacityformat`
  - Enum suffix stripping: `WindowCovering.TypeEnum` → `Type`, `WindowCovering.EndProductTypeEnum` → `EndProductType`, `ModeTagEnum` → `ModeTag`
  - Bitmap suffix stripping: `WindowCovering.ModeBitmap` → `Mode`, `ConfigStatusBitmap` → `ConfigStatus`, etc.
  - DoorLock legacy names: `LockStateEnum` → `DlLockState`, `LockTypeEnum` → `DlLockType`, `StatusCodeEnum` → `DlStatus`
  - Cluster-specific StatusCode handling: stripped to `StatusCode` in AdministratorCommissioning/RvcCleanMode/RvcRunMode/TimeSynchronization, kept as `StatusCodeEnum` elsewhere
  - Custom cluster acronym prevention: `DraftElectricalMeasurementCluster` uses `Ac`/`Rms` (not `AC`/`RMS`)
- **`K_VALUE_OVERRIDES`**: Enum/bitmap member name overrides:
  - `HueSaturation` → `HueAndSaturation`, `CredentialOverTheAirAccess` → `CredentialsOverTheAirAccess`
  - `TiltBlindLift` → `TiltBlindLiftAndTilt`, `Unbolting` → `Unbolt`
  - `LedFeedback` → `LedFeedback` (prevent LED acronym expansion in WindowCovering)
- **`FIELD_NAME_OVERRIDES`**: Additional field name fixes:
  - `Lqi` → `lqi` (CHIP SDK uses lowercase in NetworkCommissioning/ThreadNetworkDiagnostics struct fields)
- **`EXTRA_ENUM_MEMBERS`**: Inject enum members present in the CHIP SDK but missing from the Matter.js model. Used for `DoorLock.DlStatus` which needs `kSuccess`, `kFailure`, `kInvalidField`, `kResourceExhausted`, `kNotFound`.
- **`EXTRA_BITMAPS`**: Inject entire bitmap definitions missing from the Matter.js model. Used for `ColorControl.Bitmaps.ColorCapabilitiesBitmap` (`kHueSaturation`, `kEnhancedHue`, `kColorLoop`, `kXy`, `kColorTemperature`).
- **Cluster-qualified `FIELD_NAME_OVERRIDES`**: `toCamelCase()` now accepts optional cluster context for cluster-specific field name overrides (e.g., `rmsVoltage` only in DraftElectricalMeasurementCluster).

### Comparison scripts

Added two utility scripts under `python_client/scripts/` for verifying compatibility. These are rough developer tools used during this work and can be dropped later:
- `compare_clusters.py`: Detailed per-cluster structural comparison between generated and CHIP SDK clusters
- `compare_summary.py`: Categorized summary of systematic differences (detects case, Enum suffix, and Bitmap suffix mismatches)

### Intentionally not changed

These differences from the older CHIP SDK were identified but deliberately kept as-is, since the generated output is more correct or forward-compatible:

- **`eventList` global attribute**: The CHIP SDK version doesn't include `eventList` (tag 0xFFFA), but it is part of the Matter spec and newer CHIP SDK versions include it. Kept.
- **`kUnknownEnumValue` calculation**: The older CHIP SDK sometimes uses 0, but the current approach (first unused value after the max defined value) avoids assigning semantically meaningful low values. Kept.
- **Bitmap types in annotations**: The older CHIP SDK uses plain `uint` for bitmap-typed fields, but using the actual bitmap type (e.g., `OptionsBitmap`) is more type-safe. Kept.

### Issues to fix upstream in Matter.js

These are data gaps or bugs in the Matter.js model that required workarounds in the generator. Fixing them upstream would allow removing the corresponding overrides:

- **Duplicate `MinLevel` attribute in LevelControl**: The model defines `MinLevel` (id 2) twice — once for the Lighting conformance variant and once for non-Lighting. The generator deduplicates by ID, but the model should only expose one.
- **Empty attribute stubs in derived clusters**: Clusters derived from a base (e.g., DishwasherMode from ModeBase) have local attribute stubs with no type or children that shadow the base definition. This caused `supportedModes` to lose its `List[ModeOptionStruct]` type. The generator now falls back to the base, but ideally the model would inherit the full definition.
- **Missing `ColorCapabilitiesBitmap` in ColorControl**: The CHIP SDK defines this bitmap but it's absent from the Matter.js model. Currently injected via `EXTRA_BITMAPS`.
- **Missing `DlStatus` enum members in DoorLock**: The `StatusCodeEnum` only has `Duplicate` and `Occupied`, but the CHIP SDK's `DlStatus` also includes `kSuccess`, `kFailure`, `kInvalidField`, `kResourceExhausted`, `kNotFound`. Currently injected via `EXTRA_ENUM_MEMBERS`.
- **`TiltBlindLift` naming in WindowCovering.TypeEnum**: Matter.js uses `TiltBlindLift` (value 8) but the CHIP SDK calls it `TiltBlindLiftAndTilt`. Currently handled via `K_VALUE_OVERRIDES`.
- **`Unbolting` naming in DoorLock Feature bitmap**: Matter.js uses `Unbolting` but the CHIP SDK uses `Unbolt`. Currently handled via `K_VALUE_OVERRIDES`.
- **Attributes with undefined `type` but valid `metatype`**: Custom cluster attributes (e.g., EveCluster `pressure`, `voltage`, `altitude`) and some inherited attributes (BridgedDeviceBasicInformation `vendorName`, `reachable`) have `type: undefined` in the model. The generator now falls back to `metatype` (float → float32, string → str, boolean → bool), but ideally the model would set the `type` field explicitly.
- **Channel.ProgramStruct.externalIDList wrong struct reference**: Matter.js references `ContentLauncher.AdditionalInfoStruct` but CHIP SDK uses `Channel.ProgramCastStruct`. Cross-cluster reference issue in the model.
- **Descriptor.tagList uses `Globals.Structs.semtag`**: CHIP SDK uses `Descriptor.Structs.SemanticTagStruct`. The struct is defined globally in Matter.js but locally in the CHIP SDK.
- **IlluminanceMeasurement.lightSensorType resolves to `uint`**: Should be `LightSensorTypeEnum` but the enum reference isn't resolved. Model-level type resolution issue.
- **ContentAppObserver.ContentAppMessage `data`/`encodingHint` optionality swapped**: `data` is optional in CHIP SDK but mandatory in generated; `encodingHint` is the reverse. Conformance mismatch in the model.

### Issues to address in HA

- **Pylint `E0611` (no-name-in-module) for `chip.clusters.Objects`**: Now resolved by renaming the per-cluster directory from `objects/` to `cluster_defs/`, which eliminates the macOS case-insensitivity collision that caused pylint to read `objects/__init__.py` instead of `Objects.py`.

### Remaining known differences (not fixable at generator level)

- **Module path differences** (~284): Expected from the split-file architecture (`chip.clusters.cluster_defs.X.X` vs `chip.clusters.Objects.X`)
- **`aenum` vs `enum`** (~189): Inherent difference in `MatterIntEnum` base class between CHIP SDK and our implementation
- **Spec version differences**: Some enums/bitmaps/structs/commands exist in only one side due to different Matter spec versions between Matter.js and the CHIP SDK
- **Optionality mismatches** (~79 total across descriptors, commands, structs, events): Some fields are optional in CHIP SDK but mandatory in generated (or vice versa), due to conformance differences in the Matter.js model vs CHIP SDK. Notable: BridgedDeviceBasicInformation fields, mode cluster `statusText`, MediaPlayback `trackAttributes` nullability.
- **Cross-cluster struct references**: Matter.js resolves some structs from different clusters than the CHIP SDK (e.g., `Channel.ProgramStruct.externalIDList` references `ContentLauncher.AdditionalInfoStruct` instead of `Channel.ProgramCastStruct`; `Descriptor.tagList` uses `Globals.Structs.semtag` instead of `Descriptor.Structs.SemanticTagStruct`)
- **Field name casing** (a few remaining): `GroupKeyManagement.groupId` vs `groupID`, `ThreadNetworkDirectory.extendedPanId` vs `extendedPanID` — trade-offs where the override benefits the majority of uses

## Test plan

- [x] Generator runs without errors (`npx tsx python_client/scripts/generate-python-clusters.ts`)
- [x] Comparison script confirms 0 duplicate fields, 0 name mismatches
- [x] LevelControl cluster matches CHIP SDK structurally (verified manually)
- [x] HA Matter integration loads without import errors (WindowCovering, DoorLock, DraftElectricalMeasurementCluster, ElectricalPowerMeasurement, device_types verified)
- [x] Mode cluster `supportedModes` correctly typed as `List[ModeOptionStruct]` (DishwasherMode, RvcRunMode, etc.)
- [x] EveCluster float attributes (`pressure`, `voltage`, etc.) correctly typed as `float32`
- [x] Temperature/power/energy attributes correctly typed as `int` (Thermostat, TemperatureMeasurement, DeviceEnergyManagement)
- [x] Pylint E0611 resolved after `objects/` → `cluster_defs/` rename
- [x] DoorLock `requirePINforRemoteOperation` field name correct (lowercase `for`)
- [x] GeneralDiagnostics `IPv4Addresses`/`offPremiseServicesReachableIPv4` field names correct
- [x] Thermostat calibration/deadband/changeAmount correctly typed as `int` (signed)
- [x] NetworkCommissioning/ThreadNetworkDiagnostics `lqi` field name correct (lowercase)
- [x] WindowCovering `kLedFeedback` bitmap member correct (not `kLEDFeedback`)
- [x] Full comparison run: 0 duplicate fields, 0 name mismatches, 0 empty commands
- [x] Filtered comparison: 284 remaining differences are all model-level (optionality, spec version, cross-cluster references)
- [ ] Full HA integration test with commissioned devices

### DIFF
--- a/python_client/matter_server/client/client.py
+++ b/python_client/matter_server/client/client.py
@@ -247,7 +247,7 @@ class MatterClient:
                 fabric_id=f.fabricID,
                 vendor_id=f.vendorID,
                 fabric_index=f.fabricIndex,
-                fabric_label=f.label if f.label else None,
+                fabric_label=f.label or None,
                 vendor_name=vendors_map.get(str(f.vendorID)),
             )
             for f in fabrics
--- a/python_client/matter_server/client/models/node.py
+++ b/python_client/matter_server/client/models/node.py
@@ -357,7 +357,7 @@ class MatterNode:
 
     def update_attribute(self, attribute_path: str, new_value: Any) -> None:
         """Handle Attribute value update."""
-        endpoint_id = int(attribute_path.split("/")[0])
+        endpoint_id = int(attribute_path.split("/", maxsplit=1)[0])
         if endpoint_id not in self.endpoints:
             # race condition when a bridge is in the process of adding a new endpoint
             return


TheJulianJES:
- https://github.com/matter-js/matterjs-server/pull/303 not only modifies the CHIP cluster definitions but `matter_server` models as well, like device types. These are incompatible as well.
  - We should either drop them for that PR or fix them here as well.
    - I'll try to fix them here as well for now, so I can continue testing.
    - It would be cleaner to maybe have those auto-generation changes in another PR though.
  - E.g. `HeatingCoolingUnit` seems to be missing completely now (used by HA [here](https://github.com/home-assistant/core/blob/e80caaa7cd46080dc35ecf624eb24cd8964566c6/homeassistant/components/matter/switch.py#L209))
    - Does this even exist in Matter.js?
  - There also appear to be other unrelated changes in that PR to `matter_server` that seem like weird AI generations, e.g. why is a separate `msg` str for every exception created before now?

TheJulianJES:
Uhh, I'll have to check if this is accurate..: But we have both a `chip/clusters/objects/` folder and `Objects.py` module (since the original generation PR). When pylint looks for `chip.clusters.Objects`, it resolves to the folder. Everything else (correctly) resolves to the `Object.py` module though. 

The cleanest fix is to just rename the `objects/` folder to something else. It'll make the diff to the other PR a bit bigger and require some more generator changes, but it's all auto-generated anyway, so I'll do that for now. With that, pylint in HA also passes.

TheJulianJES:
PR description still needs to be updated. And other differences to look at later:

---

Critical (would cause wrong values or crashes in HA):

1. Temperature fields as uint instead of int — Thermostat setpoints, TemperatureMeasurement. Negative temperatures would be wrong. This affects the metatype: "integer" fallback — when type is undefined, signed integers also default to uint.
2. DeviceEnergyManagement/ElectricalEnergyMeasurement signed int vs uint — Power/energy values that can be negative.
3. BridgedDeviceBasicInformation optionality inversions — Fields mandatory in Gen but optional in CHIP (would crash if device doesn't provide them).
4. Mode clusters statusText mandatory vs optional — Across all mode clusters.
5. MediaPlayback TrackStruct.trackAttributes lost nullability — Would crash on null.

Field name mismatches (could cause attribute lookup failures):

6. GeneralDiagnostics IPv4Addresses vs IPV4Addresses — We have FIELD_NAME_OVERRIDES but they may not be applied to struct field contexts.
7. GroupKeyManagement groupId vs groupID — Known trade-off.
8. DoorLock requirePINforRemoteOperation — We have override, may not apply everywhere.
9. ThreadNetworkDirectory extendedPanID vs extendedPanId

Structural issues:

10. Channel ProgramStruct.externalIDList — References wrong cluster struct (ContentLauncher.AdditionalInfoStruct instead of ProgramCastStruct)
11. Descriptor tagList — Uses Globals.Structs.semtag instead of Descriptor.Structs.SemanticTagStruct
12. WindowCovering commands have extra ignored field

python_client/scripts/generate-python-clusters.ts
```diff
@@ -252,6 +311,133 @@ const FIELD_NAME_OVERRIDES: Record<string, string> = {
...
+            members: [
+                { kName: "kHueSaturation", value: 0x1 },
+                { kName: "kEnhancedHue", value: 0x2 },
+                { kName: "kColorLoop", value: 0x4 },
+                { kName: "kXy", value: 0x8 },
+                { kName: "kColorTemperature", value: 0x10 },
+            ],
+        },
+    ],
+};
```
TheJulianJES, line 439: These are needed by HA. We should look at if they should be added to Matter.js.

python_client/scripts/generate-python-clusters.ts
```diff
@@ -1522,6 +1797,43 @@ function generateDeviceTypes(): string {
...
+        w.line(`device_type: int = 0x${extra.id.toString(16).toUpperCase().padStart(4, "0")}`);
+        w.line("clusters: set[type[Cluster]] = {");
+        w.pushIndent();
+        for (const ref of extra.clusters.sort()) {
+            w.line(`${ref},`);
+        }
+        w.popIndent();
+        w.line("}");
+        w.popIndent();
+    }
```
TheJulianJES: `HeatingCoolingUnit` was a provisional device type, being completely removed in Matter device library specifications v1.3. It is currently used by HA for matching a switch entity, but no certified devices should use it.

With the linked PR below, HA would no longer use this, so we could then remove adding it here:
- https://github.com/home-assistant/core/pull/166828

python_client/scripts/generate-python-clusters.ts
Apollon77: Ok Statuscode are "just" limited copies from the Global StatusCode table and is in no spec, when the SDK added that why ever. We can leave for backward compatibility or just adjust in HA to use the global StatusCode table -that would be better for the future anyway

ColorCapabilitiesBitmap .. ok I found the reason why the enum is missing. I can add for 1.5 next update, for now we can leave that in OR:
> Bits 0-4 of the ColorCapabilities attribute SHALL have the same values as the corresponding bits of
the FeatureMap attribute. All other bits in ColorCapabilities SHALL be 0.

... replace by feature lookup :-)

python_client/scripts/generate-python-clusters.ts
TheJulianJES: As the above HA PR is merged, the device type is removed now: [`21495b7` (this PR)](https://github.com/matter-js/matterjs-server/pull/457/commits/21495b79291a2bfd59eb80c741b70fc747dbffb5)

PR #513, opened 2026-04-08
Fix node availability events not emitted on `Disconnected` state

### LINKED ISSUE #494: Matter.js does not communicate "unavailable" state to HA
### Version

v0.5.13

### Node.js Version

node@22.21.1 | linux | x64

### Operating System

Linux (x64)

### Issue Description

t appears to be that HA Core is not getting a message that it "understands" to mark a device as "Unavailable" when a device is declare as offline in matter.js. I'm unsure whether this is a HA Core or matter.js issue.

At 10:55 I pulled the power to a switch (Node @1:12 in this example). Within a few minutes the Node has been declared “offline” by matter.js however HA still shows the device as available. I let this run for 15 mins.... no change, the device is unplugged, matter.js sees it as offline but HA sees it as online.

The logs (just filtered for that device) were as follows:

```
2026-04-06 10:50:36.272 INFO   WebSocketC~erHandler [0] Node @1:12 availability changed to true
2026-04-06 10:51:58.368 INFO   ClientInteraction    Invoke » @1:12•ce23⇵11ca 0x1.OnOff:0x6.off:0x0 with (no payload)
2026-04-06 10:51:58.426 INFO   ClientInteraction    Invoke « @1:12•ce23⇵11ca 0x1.OnOff:0x6.off:0x0 status: Success (0)
2026-04-06 10:51:59.796 INFO   ClientInteraction    Invoke » @1:12•ce23⇵11cb 0x1.OnOff:0x6.on:0x1 with (no payload)
2026-04-06 10:51:59.854 INFO   ClientInteraction    Invoke « @1:12•ce23⇵11cb 0x1.OnOff:0x6.on:0x1 status: Success (0)
2026-04-06 10:52:04.503 INFO   ClientEventEmitter   Received event switch.initialPress on server-2-134b.@1:12.ep1.ep4 newPosition: 1
2026-04-06 10:52:05.537 INFO   ClientEventEmitter   Received event switch.shortRelease on server-2-134b.@1:12.ep1.ep4 previousPosition: 1
2026-04-06 10:52:05.549 INFO   ClientEventEmitter   Received event switch.multiPressComplete on server-2-134b.@1:12.ep1.ep4 previousPosition: 0 totalNumberOfPressesCounted: 1
2026-04-06 10:53:03.947 INFO   ClientInteraction    Invoke » @1:12•ce23⇵11cf 0x1.OnOff:0x6.on:0x1 with (no payload)
2026-04-06 10:53:04.007 INFO   ClientInteraction    Invoke « @1:12•ce23⇵11cf 0x1.OnOff:0x6.on:0x1 status: Success (0)
2026-04-06 10:53:53.584 INFO   ClientEventEmitter   Received event switch.initialPress on server-2-134b.@1:12.ep1.ep5 newPosition: 1
2026-04-06 10:53:54.599 INFO   ClientEventEmitter   Received event switch.shortRelease on server-2-134b.@1:12.ep1.ep5 previousPosition: 1
2026-04-06 10:53:54.610 INFO   ClientEventEmitter   Received event switch.multiPressComplete on server-2-134b.@1:12.ep1.ep5 previousPosition: 0 totalNumberOfPressesCounted: 1
2026-04-06 10:56:35.605 INFO   ClientSubscription   Subscription c9e931e2 to peer @1:12 timed out after 1m 40s
2026-04-06 10:56:35.607 ERROR  ClientSubscription   Replacing subscription to @1:12 due to timeout
2026-04-06 10:56:53.476 INFO   ClientInteraction    Read » @1:12•ce23⇵11d1 attributes: 0x0/OperationalCredentials(0x3e)/Fabrics(0x1) fabricFiltered: false
2026-04-06 10:56:53.577 INFO   ClientInteraction    Read » @1:12•ce23⇵11d2 attributes: 0x0/OperationalCredentials(0x3e)/Fabrics(0x1) fabricFiltered: false
2026-04-06 10:57:18.212 INFO   ClientNode           server-2-134b.@1:12 is offline
2026-04-06 10:57:18.214 INFO   Session              @1:12•ce23 Session ended
2026-04-06 10:57:18.216 ERROR  ClientSubscription   Failed to probe reachability of peer @1:12, resubscribe with new session
2026-04-06 10:57:18.216 INFO   IpServiceStatus      @1:12 Resolving (address is unreachable)
2026-04-06 10:57:18.226 INFO   PeerConnection       @1:12•unsecured#1e43761242180
```
If we look at the Python server same routine, same switch, pull the power (at approx: 11:15) and wait we get this:

```
2026-04-06 11:16:49.503 (Dummy-2) CHIP_ERROR [chip.native.DMG] Subscription Liveness timeout with SubscriptionID = 0xb2fee7ad, Peer = 01:0000000000000012
2026-04-06 11:16:49.508 (MainThread) INFO [matter_server.server.device_controller] <Node:18> Subscription failed with CHIP Error 0x00000032: Timeout, resubscription attempt 0
2026-04-06 11:17:13.558 (Dummy-2) CHIP_ERROR [chip.native.EM] <<5 [E:4402i with Node: <0000000000000000, 0> S:0 M:58331174] (U) Msg Retransmission to 0:0000000000000000 failure (max retries:4)
2026-04-06 11:17:24.125 (Dummy-2) CHIP_ERROR [chip.native.SC] CASESession timed out while waiting for a response from peer <0000000000000012, 1>. Current state was 4
2026-04-06 11:17:24.126 (Dummy-2) CHIP_ERROR [chip.native.DMG] Failed to establish CASE for re-subscription with error 'src/protocols/secure_channel/CASESession.cpp:594: CHIP Error 0x00000032: Timeout'
2026-04-06 11:17:24.128 (MainThread) INFO [matter_server.server.device_controller] <Node:18> Subscription failed with CHIP Error 0x00000032: Timeout, resubscription attempt 1
2026-04-06 11:17:55.461 (Dummy-2) CHIP_ERROR [chip.native.EM] <<5 [E:4403i with Node: <0000000000000000, 0> S:0 M:58331175] (U) Msg Retransmission to 0:0000000000000000 failure (max retries:4)
2026-04-06 11:18:06.066 (Dummy-2) CHIP_ERROR [chip.native.SC] CASESession timed out while waiting for a response from peer <0000000000000012, 1>. Current state was 4
2026-04-06 11:18:06.067 (Dummy-2) CHIP_ERROR [chip.native.DMG] Failed to establish CASE for re-subscription with error 'src/protocols/secure_channel/CASESession.cpp:594: CHIP Error 0x00000032: Timeout'
2026-04-06 11:18:06.068 (MainThread) INFO [matter_server.server.device_controller] <Node:18> Subscription failed with CHIP Error 0x00000032: Timeout, resubscription attempt 2
2026-04-06 11:18:06.069 (MainThread) INFO [matter_server.server.device_controller] <Node:18> Marked node as unavailable
```
As soon as the “Marked node as unavailable” event turns up in the second example the device shows as unavailable in HA. My guess (note guess) is that HA is looking for something similar to “Node @1:12 availability changed to false” from matter.js which it never sees and hence the device appears to be still online in the frontend when it is not.

### Steps to Reproduce

Verify Thread/Matter Device is connected to HA and can be controlled by it and is subscribed to events.
Remove power from device and observe matter logs and HA frontend for the device.

### Server Logs

```shell
See above
```

### Device Information

Device Inovelli VTM31-SN 

### Additional Context

_No response_

### Pre-submission Checklist

TheJulianJES:
Like this PR, the description below is AI-generated but still worth a read.
Feel free to push changes directly, or close this PR and open another one that attempts a similar (or different) fix.

Fixes https://github.com/matter-js/matterjs-server/issues/494

### Problem

When a Matter device is unplugged or becomes unreachable, the Matter.js server dashboard correctly shows it as offline, but Home Assistant never marks the device as unavailable. The device remains shown as available indefinitely in HA until the server is restarted.

### Root cause

This is a regression introduced in #162 (82068d1), which added the availability tracking and debouncing system. That commit introduced both the `nodeAvailabilityChanged` event (the new mechanism for pushing availability updates to WebSocket clients) and an early return guard in the `stateChanged` handler that skips all processing when a node enters the `Disconnected` state:

```typescript
node.events.stateChanged.on(state => {
    if (state === NodeStates.Disconnected) {
        return; // Blocks ALL processing below
    }
    // ... availability calculation, state tracking, event emission
});
```

Before #162, availability worked because:
1. `nodeStateChanged` was emitted for all states including `Disconnected`
2. `available` was computed live as `node.isConnected` each time node data was serialized

After #162, the approach changed to a push-based model using `nodeAvailabilityChanged`, and `nodeStateChanged` was deliberately changed to no longer trigger `node_updated` events (the comment reads: *"Availability changes are handled by nodeAvailabilityChanged"*). However, the `Disconnected` guard prevents the new `nodeAvailabilityChanged` event from ever firing for that state — breaking both the new path and the old one simultaneously.

The guard also caused a secondary bug: `previousState` was never updated for `Disconnected` transitions. This meant the debounce logic could produce incorrect results. For example, in a Connected → Disconnected → Reconnecting sequence, the debounce would see `previousState = Connected` (stale) and incorrectly report the node as still available.

The availability tracking system introduced in #162 shipped with no test coverage, which is why this regression went undetected. The `ws-controller` package had no tests at all. Broader test coverage for ws-controller would be worth addressing in future PRs.

### Fix

Remove the `Disconnected` guard so all state transitions are processed uniformly. The availability tracking logic already handles `Disconnected` correctly — `isNodeAvailable()` returns `false` for it — it was just never being called.

Additionally, move the availability tracking state (`#lastAvailability` map) and the state-change processing logic from `ControllerCommandHandler` into `Nodes`, where the related state (`#previousStates`) and availability logic (`isNodeAvailable`) already live. This consolidates the availability concern into a single class via a new `processStateChange()` method that:
1. Computes availability from the new state and previous state
2. Updates internal tracking maps
3. Returns whether availability actually changed (and the new value)

`ControllerCommandHandler` now calls `processStateChange()` and reacts to the result, rather than managing the tracking maps directly.

### Commits (reviewable individually)

1. **Fix node availability not updating on Disconnected state** — the bugfix: removes the `Disconnected` early return (4 lines deleted)
2. **Move availability tracking logic from ControllerCommandHandler into Nodes** — refactor, also fixes `#lastAvailability` entries not being cleaned up when a node is removed
3. **Add unit tests for Nodes availability logic** — 14 tests + test infrastructure

Each commit is independent — commits 2 and/or 3 can be dropped if the refactor or tests are unwanted. Note that dropping commit 2 leaves the `#lastAvailability` cleanup missing from node removal — a minor memory leak that existed before this PR.

### Known pre-existing issue: debounce inconsistency between event and serialization paths

There are two paths that determine node availability, and they can disagree during the `Reconnecting` debounce window:

1. **Event path** (`processStateChange`): Reads `previousState` *before* updating it. For a Connected → Reconnecting transition, it sees `(Reconnecting, Connected)` → available (debounce). No `nodeAvailabilityChanged` event is emitted.
2. **Serialization path** (`isAvailable`, called by `getNodeDetails`): Reads `node.connectionState` live from the `PairedNode` and `previousState` from `#previousStates` — which has *already been updated* to `Reconnecting`. It sees `(Reconnecting, Reconnecting)` → unavailable.

This means that during the debounce window, if a `node_updated` event is triggered for an unrelated reason (e.g., a structure change or attribute update), the serialized node data will include `available: false` even though no `nodeAvailabilityChanged` event was emitted.

This inconsistency predates this PR — the same read-then-update vs read-after-update ordering existed in the original inline code in `ControllerCommandHandler`. The proper fix would be to have `isAvailable()` return the cached `#lastAvailability` value instead of recomputing from the state pair, but that is a separate change.

### Further AI summary on changes

<details><summary>Furhther AI summary (click to open)</summary>
<p>

### Changes

- **`Nodes.ts`**: Add `#lastAvailability` map, `seedState()` method for initial state setup, and `processStateChange()` method that encapsulates the full state transition logic. Clean up `#lastAvailability` in `delete()`.
- **`ControllerCommandHandler.ts`**: Remove `#lastAvailability` map (moved to `Nodes`). Replace inline state tracking logic in the `stateChanged` handler with a call to `nodes.processStateChange()`. Replace manual state seeding with `nodes.seedState()`.
- **`test/NodesTest.ts`** (new): 14 unit tests covering `isNodeAvailable()` and `processStateChange()`. The `processStateChange` tests cover the exact scenario that was broken (Connected → Disconnected must report `availabilityChanged: true`) and the full device lifecycle (Connected → Reconnecting → Disconnected → Connected). Re-adding the original `Disconnected` guard causes 3 test failures.
- **`ws-controller/package.json`**, **`ws-controller/tsconfig.json`**, **`ws-controller/test/tsconfig.json`** (new): Test infrastructure for ws-controller, which previously had no tests.

### How it was verified

- Confirmed that Matter.js `PairedNode.#setConnectionState()` already deduplicates state transitions (returns early if state is unchanged), so there is no risk of event floods
- Verified the two non-device-offline sources of `Disconnected` are harmless:
  - **Decommissioning**: `Disconnected` fires before `decommissioned` — HA gets a `node_updated(available: false)` followed by `node_removed`, which is correct
  - **Controller shutdown**: WebSocket connection is tearing down, events have no effect
- Verified that nodes starting in `Disconnected` state (the Matter.js default) don't produce spurious availability events: `seedState` defaults `lastAvailability` to `false`, matching the `Disconnected` → unavailable calculation
- All existing tests pass (299 tests across ws-controller, ws-client, and matter-server)

</p>
</details>

### DIFF
--- a/packages/ws-controller/package.json
+++ b/packages/ws-controller/package.json
@@ -16,7 +16,8 @@
     "scripts": {
         "clean": "matter-build clean",
         "build": "matter-build",
-        "build-clean": "matter-build --clean"
+        "build-clean": "matter-build --clean",
+        "test": "matter-test --force-exit"
     },
     "engines": {
         "node": ">=20.19.0 <22.0.0 || >=22.13.0"
--- a/packages/ws-controller/src/controller/ControllerCommandHandler.ts
+++ b/packages/ws-controller/src/controller/ControllerCommandHandler.ts
@@ -307,10 +307,6 @@ export class ControllerCommandHandler {
         });
         node.events.eventTriggered.on(data => this.events.eventChanged.emit(nodeId, data));
         node.events.stateChanged.on(state => {
-            if (state === NodeStates.Disconnected) {
-                return;
-            }
-
             // Calculate availability before and after state change
             const previousState = this.#nodes.getPreviousState(nodeId);
             const wasAvailable = this.#lastAvailability.get(nodeId) ?? false;
--- a/packages/ws-controller/src/controller/ControllerCommandHandler.ts
+++ b/packages/ws-controller/src/controller/ControllerCommandHandler.ts
@@ -160,8 +160,6 @@ export class ControllerCommandHandler {
     #availableUpdates = new Map<NodeId, SoftwareUpdateInfo>();
     /** Poller for custom cluster attributes (Eve energy, etc.) */
     #customClusterPoller: CustomClusterPoller;
-    /** Track the last known availability for each node to detect changes */
-    #lastAvailability = new Map<NodeId, boolean>();
     /** Track in-flight invoke-commands for deduplication across all WebSocket connections */
     readonly #inFlightInvokes = new Map<string, Promise<unknown>>();
     events = {
@@ -307,11 +305,6 @@ export class ControllerCommandHandler {
         });
         node.events.eventTriggered.on(data => this.events.eventChanged.emit(nodeId, data));
         node.events.stateChanged.on(state => {
-            // Calculate availability before and after state change
-            const previousState = this.#nodes.getPreviousState(nodeId);
-            const wasAvailable = this.#lastAvailability.get(nodeId) ?? false;
-            const isAvailable = this.#nodes.isNodeAvailable(state, previousState);
-
             // Only refresh cache on Connected state (not Reconnecting, WaitingForDiscovery, etc.)
             if (state === NodeStates.Connected) {
                 attributeCache.update(node);
@@ -322,19 +315,18 @@ export class ControllerCommandHandler {
                 }
             }
 
-            // Update state tracking for the next transition
-            this.#nodes.setPreviousState(nodeId, state);
-            this.#lastAvailability.set(nodeId, isAvailable);
+            // Process state change and check if availability changed
+            const result = this.#nodes.processStateChange(nodeId, state);
 
             // Emit state changed event
             this.events.nodeStateChanged.emit(nodeId, state);
 
             // Emit availability changed if it actually changed
-            if (wasAvailable !== isAvailable) {
+            if (result.availabilityChanged) {
                 logger.info(
-                    `Node ${this.formatNode(nodeId)} availability changed: ${wasAvailable} -> ${isAvailable} (state: ${NodeStates[previousState ?? -1]} -> ${NodeStates[state]})`,
+                    `Node ${this.formatNode(nodeId)} availability changed to ${result.available} (state: ${NodeStates[state]})`,
                 );
-                this.events.nodeAvailabilityChanged.emit(nodeId, isAvailable);
+                this.events.nodeAvailabilityChanged.emit(nodeId, result.available);
             }
         });
         node.events.structureChanged.on(() => {
@@ -354,12 +346,7 @@ export class ControllerCommandHandler {
 
         // Seed state tracking from the node's current connection state so the first
         // stateChanged event shows a real previous state instead of "undefined".
-        const initialState = node.connectionState;
-        this.#nodes.setPreviousState(nodeId, initialState);
-        // Only mark as available when actually connected; all other states default to false (unavailable).
-        if (initialState === NodeStates.Connected) {
-            this.#lastAvailability.set(nodeId, true);
-        }
+        this.#nodes.seedState(nodeId, node.connectionState);
 
         // Initialize attribute cache if node is already initialized
         if (node.initialized) {
--- a/packages/ws-controller/src/controller/Nodes.ts
+++ b/packages/ws-controller/src/controller/Nodes.ts
@@ -27,6 +27,8 @@ export class Nodes {
     #attributeCache = new AttributeDataCache();
     /** Track previous connection state for availability debouncing */
     #previousStates = new Map<NodeId, NodeStates>();
+    /** Track the last known availability for each node to detect changes */
+    #lastAvailability = new Map<NodeId, boolean>();
 
     /**
      * Get the attribute cache instance.
@@ -75,6 +77,7 @@ export class Nodes {
         this.#nodes.delete(nodeId);
         this.#attributeCache.delete(nodeId);
         this.#previousStates.delete(nodeId);
+        this.#lastAvailability.delete(nodeId);
     }
 
     /**
@@ -127,6 +130,42 @@ export class Nodes {
         const currentState = node.connectionState;
         const previousState = this.#previousStates.get(nodeId);
         return this.isNodeAvailable(currentState, previousState);
+    }
+
+    /**
+     * Seed initial state tracking for a newly registered node.
+     * Only marks as available when actually connected; all other states default to unavailable.
+     */
+    seedState(nodeId: NodeId, initialState: NodeStates): void {
+        this.#previousStates.set(nodeId, initialState);
+        if (initialState === NodeStates.Connected) {
+            this.#lastAvailability.set(nodeId, true);
+        }
+    }
+
+    /**
+     * Process a node state change and return whether availability changed.
+     * Updates internal state tracking and returns the availability transition if one occurred.
+     *
+     * @returns Object with `availabilityChanged: true` and `available` value if availability changed,
+     *          or `availabilityChanged: false` if it did not.
+     */
+    processStateChange(
+        nodeId: NodeId,
+        newState: NodeStates,
+    ): { availabilityChanged: true; available: boolean } | { availabilityChanged: false } {
+        const previousState = this.#previousStates.get(nodeId);
+        const wasAvailable = this.#lastAvailability.get(nodeId) ?? false;
+        const isAvailable = this.isNodeAvailable(newState, previousState);
+
+        // Update state tracking for the next transition
+        this.#previousStates.set(nodeId, newState);
+        this.#lastAvailability.set(nodeId, isAvailable);
+
+        if (wasAvailable !== isAvailable) {
+            return { availabilityChanged: true, available: isAvailable };
+        }
+        return { availabilityChanged: false };
     }
 
     /**
--- a/packages/ws-controller/test/NodesTest.ts
+++ b/packages/ws-controller/test/NodesTest.ts
@@ -1,0 +1,146 @@
+/**
+ * @license
+ * Copyright 2025-2026 Open Home Foundation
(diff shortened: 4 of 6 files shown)


Apollon77:
Good analysis and clean code, but the root cause diagnosis is incorrect — `Disconnected` is not reached when devices go offline.

### What actually happens when a device goes offline

Looking at the debug logs attached to #494, when a device is unplugged:

1. `Connected → Reconnecting` (subscription goes inactive, MDNS address still cached)
2. The node **stays in `Reconnecting` indefinitely** — it never reaches `WaitingForDeviceDiscovery` because `peer.service.addresses.size` remains > 0 (cached MDNS address)
3. `Disconnected` is **never reached** — it's only set by `decommission()` and `close()` in matter.js `PairedNode`

The debug log from #494 confirms: `PeerConnection @1:12 Connecting addr #: 1` — the address stays cached, the node retries Sigma1 for ~9 minutes in `Reconnecting` state, never transitioning to `WaitingForDeviceDiscovery` or `Disconnected`.

### The real bugs

**Bug 1 — Event/serialization path split-brain:**
The `stateChanged` handler reads `previousState` *before* updating it → `isNodeAvailable(Reconnecting, Connected) = true` (debounce) → no `nodeAvailabilityChanged` event emitted. But `isAvailable()` (called by `getNodeDetails()` for WS serialization) reads `previousState` *after* update → `isNodeAvailable(Reconnecting, Reconnecting) = false`.

Result: HA (connected via WS, relies on events) never gets notified. The dashboard (reconnects WS, gets fresh `start_listening` response with live `isAvailable()`) correctly shows offline.

**Bug 2 — Unbounded Reconnecting debounce:**
Even if Bug 1 were fixed, a device stuck in `Reconnecting` with cached MDNS addresses would stay "available" forever because the debounce has no timeout.

### Our fix

1. `isAvailable()` returns the **cached** availability decision from the last `processStateChange()` call, eliminating the split-brain
2. A **5-minute reconnect timer** per node: if a node stays in `Reconnecting` for 5 minutes, it's forced unavailable
3. The `Disconnected` guard is also removed (same as this PR) — not because it fixes the offline case, but because it prevented state tracking from updating on decommission

PR with the fix coming shortly.

TheJulianJES:
Thanks! This is superseded by:
- https://github.com/matter-js/matterjs-server/pull/515/

That PR looks good to me and fixes all issues described in the description here properly.
EDIT: There seems to be some other issue with that as well that's being looked at.

PR #556, opened 2026-04-20
Add hide options for Thread graph

AlixBa:
Add a dropdown menu with several options to hide nodes or edges from a Thread graph. This can be useful for large networks with a lot of elements to display

---

Hey team,

First, thank you for the visualization of (Thread) networks, it has been very handy for me to make it more reliable. However, it starts to be very messy and I thought a few options could improve it.

I am no TypeScript developer, so I went straight to the point without adding tests as there was no existing structure for this. I tested locally with my cluster. Hope that helps!

Screenshot 2026-04-20 at 21 08 38
Screenshot 2026-04-20 at 21 08 46
Screenshot 2026-04-20 at 21 08 52
Screenshot 2026-04-20 at 21 09 03


Apollon77:
great idea, thanks, will look at it tomorrow in detail. Feel free to also add a changelog one liner in the work in progress section

Apollon77:
Thanks a lot, I will push some smaller fixups and adjustments, then  either you coulld pull and recheck it or I try to check earliest on the weekend

Apollon77:
Thanks for the contribution, @AlixBa! Pushed a fixup commit (259f44e) directly to this branch since it needed a few things before merge. Quick summary:

### Must-fix

- **Build was failing.** The new required `signalLevel` field on `ThreadConnection` wasn't set on the route-table-derived push site in `network-utils.ts` → `TS2741`. Fixed by adding a `getSignalLevelFromLqi` helper (paralleling the existing `getSignalColorFromLqi`) and populating the field. Also made `signalLevel` **optional** — route-table edges without LQI should not be classified as "weak" and accidentally hidden by the weak-signal filter; they stay visible now.
- **Accidental `@state()` on `_initialSelectionApplied`.** That field is mutated inside `updated()` and isn't meant to be reactive — would cause extra render cycles. Removed the decorator.

### Important

- **Empty-state mismatch** (flagged by Copilot review, valid): when `hideOfflineNodes` is on and all Thread nodes are offline, `_updateGraph()` clears everything but `render()` still returned the graph container. Applied the same filter in `render()` and added a distinct "No online Thread devices" empty state with a hint to disable the filter.
- **Unsafe type casts** in `_handleToggleHideOption` — `keyof MatterNetworkView` + `as boolean` is too permissive and violates our project's "avoid type casts" rule. Replaced with a tight `HideOptionKey` union and a module-scope `HIDE_OPTIONS` constant.

### Polish

- Centralized the `SignalLevel` type in `network-types.ts` and reused it in both `ThreadConnection` and `NodeConnection` instead of inlining the union.
- Dropdown outside-click now uses a document-level handler registered in `connectedCallback`/`disconnectedCallback`. The old `.content @click` only fired for clicks inside that div — clicks on header, footer, or sidebar didn't close the menu. Also added Escape-to-close.
- Added `aria-haspopup`, `aria-expanded`, `aria-controls` on the toggle button, and `role="group"` + `aria-label` on the dropdown (avoids over-claiming `role="menu"` semantics, which would require arrow-key navigation).
- Replaced the hardcoded `box-shadow: 0 2px 8px rgba(0,0,0,0.15)` with `var(--md-sys-elevation-level2, ...)` so it adapts in dark mode.
- Memoized `_getFilteredUnknownNeighbors(unknown)` in `network-details.ts` — was being called twice per render.
- Moved the new CHANGELOG Feature entry next to the other Feature line (and stripped a trailing whitespace).

### Verified locally

- `npm run format` / `npm run lint` / `npm run build` — clean
- `npm test` — 242/242 pass

Let me know if anything here reads differently than intended — happy to revise.

AlixBa:
@Apollon77 thanks for the update!

I've realized that something might not work as expected. Connections are deduplicated right now with the first visited key `from-to` or `to-from`. When I use the filter, it removes the one displayed on the graph as expected but when I want to filter the connections list in the `details` page, it's not in-sync.

**I also saw that there is a hidden property on viz.js nodes/edges that we can use instead of what I did!**

In this example I show only `medium`. The `grillplats` has 5 edges but only 4 listed connections. This is because one of the link is `strong` from `grillplats` to `x` but `medium` from `x` to `grillplats`. 
Screenshot 2026-04-21 at 14 43 12

I think what could be nice to implement is
- if there is no filter, keep the same logic as today (unless it can be changed to display the weakest signal?)
- if there is a filter and the current connection/edge is filtered, ignore this connection/edge and iterate on the next one

There is also the possibility of displaying the bidir edges but I feel like it would be too much for the eyes

Apollon77:
Let me and Claude think about that later :-)

Apollon77:
@AlixBa nice catch on the dedup mismatch, and thanks for the vis.js \`hidden\` refactor — much cleaner.

Pushed \`aa0d58e\` implementing your proposal: \`buildThreadConnections\` now takes an optional \`shouldShowConnection\` predicate. During dedup, candidates that fail the predicate \`continue\` **without** reserving the pair key, so the reverse direction gets a chance to win. If every direction of a pair is filtered, the pair is absent — which is consistent with the per-direction filter already applied by \`network-details.ts\`.

When no edge filter is active, the predicate is \`undefined\` and the behavior is identical to before (no performance impact, no direction flipping on unfiltered toggles).

In your example: "grillplats→x strong, x→grillplats medium, show only medium" — the medium direction will now always be picked regardless of node iteration order, matching what the details page shows.

The offline filter isn't in the predicate because offline state is identical for both directions of a pair, so directional preference is irrelevant there; your \`hiddenNodeIds\` set already handles it via edge endpoints.

Locally: \`format\` / \`lint\` / \`build\` clean, \`test\` 242/242.

AlixBa:
@Apollon77 Claude worked on it as I wanted to fix a few things. But that's a big PR now, :| 

- **`01044bc`** — Centralize edge computation: `buildThreadEdgePairs()` computes all directional edges (0–2 per pair), then `ThreadGraph` applies filter → dedup (keep worst signal) → save base state. Highlight swaps to the highlighted node's edge when a better-signal edge was dedup-hidden.

- **`9b23def`** — Align panel with graph: `getNodeConnectionsFromPairs()` now applies the same filter-then-dedup pipeline as the graph (with `EdgeFilterOptions`). Prefer outgoing among survivors (matches highlight swap). Panel and graph are guaranteed 1:1. Remove arrow feature (redundant now). Clean up unused imports.

- **`cc90b88`** — Fix deselect and icon color: close button calls `deselectAll()` on the graph to properly clear highlights. Node icon blue color is no longer baked at build time — it's set in `_highlightConnections` and restored in `_clearHighlights`.

edit: ha, fails on lint/build but not locally

AlixBa:
https://github.com/user-attachments/assets/d0808192-3c66-4711-9c83-542bbffc2165

Apollon77:
I have a look .. but I can not test in my home network before the weekend ... but your video is great already. SO I woud still consider to merge it

Apollon77:
Separate from the cleanup I'll push shortly (removes the CI-failing unused `_formatExtAddress`, the dead `buildThreadConnections` / `getNodeConnections` / `shouldShowConnection` predicate, a stale comment about arrows), three behavior questions surfaced during review that I'd like your read on before we merge. These are design decisions, not bugs — the code does what it's documented to do, but I'm not sure it does what users will expect.

## Background: how a Thread pair is displayed today

Each connected Thread node pair has up to two directional readings:

- **A→B**: what node A reports about its link to B, from A's neighbor table (A's own RSSI/LQI for the frames it hears from B).
- **B→A**: what node B reports about its link to A, from B's neighbor table.

These can disagree, sometimes dramatically. A high-power mains-powered router can hear a low-power sleepy end device just fine (strong) while the end device barely hears the router back (weak). That asymmetry is not noise — it's signal, and often the first clue when a mesh link is acting up.

After `3bc3114`+`de13395`, the pipeline is:

1. `buildThreadEdgePairs` builds the full set of directional edges (no dedup).
2. Each edge is filtered independently (offline cascade, signal-level hide options).
3. Among survivors per pair, a single edge is chosen for display — but the **choice rule differs between graph and panel**.

That difference is where the three questions come from.

## 1. Graph and panel can display different signal colors for the same connection before you click anything

**What's happening in the code**

- `thread-graph.ts` ~line 302: among surviving edges in a pair, `sort ascending by getEdgeSignalScore → keep index 0`. That's the **weakest** surviving edge (worst-case view).
- `network-utils.ts` `getNodeConnectionsFromPairs` ~line 1173: among surviving edges for the selected node, `find(s => s.isOutgoing)` wins if it exists, else fall back to weakest.

**Concrete scenario**

Node A is selected. The A↔B pair has A→B = strong (RSSI −60) and B→A = weak (RSSI −88). Both pass the user's filters.

- The graph draws the A–B edge in **red** (the weak B→A direction won dedup because the sort is ascending).
- The panel, showing A's perspective, lists this connection as **green** (A→B is outgoing, picked first).

Both views are internally consistent with their own rule. A user scanning the graph for trouble, then clicking a red edge to read details, will see a green RSSI number and wonder if they mis-clicked. The direction cue (`(reverse)` suffix on line 306 of `network-details.ts`) exists but isn't called out — it's easy to miss.

**Option A — match panel to graph default (weakest)**

Change `getNodeConnectionsFromPairs` to drop the outgoing preference and just use `getEdgeSignalScore` ascending, same as the graph. Consequence: the pre-click graph and panel always show the same direction and same color. The panel row labels direction honestly (it already knows `isOutgoing`).

- *Pro*: HA power users get one answer to "is this link bad?" — the worst direction, which is what will fail first.
- *Pro*: No cognitive load from two ranking systems.
- *Con*: Loses the "node-perspective" framing the panel used to have. When looking at node A, you may be reading a row about how B sees A, not how A sees B. That's arguably more useful anyway (worst-case is the failure mode that matters), but it's a semantic shift.

**Option B — keep per-node perspective, annotate when it doesn't match graph**

Keep the outgoing preference. Add a small badge/label on the panel row when the chosen edge is *not* the one drawn on the graph — e.g. a secondary line "graph shows reverse direction: RSSI −88 / LQI 120 (weak)".

- *Pro*: Preserves the "this is what node A sees" framing.
- *Pro*: Shows both directions when they disagree, which is itself diagnostic info.
- *Con*: Row becomes denser. Third data mode on top of the `(reverse)` hint.

**Option C — graph shows both directions when they disagree**

Render two parallel edges (or one split-color edge) per pair when both directions survive and differ in signal level. Most honest, but significantly more visual noise and probably needs UX work in vis.js.

My lean: **A**, because for the HA audience "is this link going to drop out?" is always the question, and the worst direction is the answer. B is also defensible if we weight the developer audience higher. C is future work.

## 2. Swap-on-click has no transition

**What's happening**

`thread-graph.ts` `_highlightConnections` ~line 438: when the highlighted node is the *remote* side of the displayed edge and a dedup-hidden edge from the highlighted node exists, we `hidden = true` on the visible edge and `hidden = false` on the dedup-hidden one, then update width/color. vis.js applies both changes on the next frame.

**What the user sees**

Click a node. The edge that was there instantly disappears and a new edge appears between a different pair of endpoints (or between the same endpoints but with different color and thickness). No fade, no movement, no indication that this is a perspective switch rather than a refresh or a bug.

**Why it matters**

For a first-time user debugging their mesh, "I clicked and stuff rearranged" is exactly the kind of thing that makes them stop trusting the diagnostic. Power users will learn the behavior but shouldn't have to.

**Options**

- Tween color/width/position on the swap (150–200 ms cubic-easing). vis.js has some animation support for color transitions; geometry changes harder.
- Keep both edges visible briefly (~300 ms) with the old one fading out, so the eye tracks the replacement.
- Add a subtle "perspective: Node A" label near the highlighted node during highlight so the user knows the view has been reframed.

Low priority. Likely a follow-up PR. Worth filing.

## 3. Arrow removal eliminates the only visible asymmetric-visibility signal

**What was removed**

`de13395` dropped the one-way arrow that showed up when the highlighted node had no outgoing edge to a peer but the peer had an outgoing edge to it. The arrow pointed toward the highlighted node and communicated "this peer can see you, but you can't see this peer back".

**Why that diagnostic matters**

Asymmetric visibility is one of the cleanest single-packet signals in Thread troubleshooting. Typical causes:

- A sleepy end device lost its parent's link but the parent's neighbor table hasn't aged the child out yet.
- TX-power mismatch (router is mains-powered with high TX, end device is battery with low TX; the strong-side hears the weak-side but not vice versa).
- Antenna orientation issue on one of the two devices.
- Range-cliff: device is at the edge of reception and dropping packets in one direction only.

Without any UI cue, these cases now look identical to healthy symmetric links. The graph coerces each pair to one direction; the panel collapses both sides into a single row (preferring outgoing or weakest, per #1). The `isOutgoing === false` signal is tracked internally but isn't surfaced in a way that makes asymmetry jump out.

**Options**

**(i) Bring the arrow back, but only during highlight.** When a node is selected and a peer has no outgoing edge from the selected node (only incoming from the peer), draw the edge with an arrowhead pointing toward the selected node. Scoped to highlight — no default-view clutter. Minimal code change: the data is still there (`pair.edgeAB` / `edgeBA` presence check), just the rendering path that was removed.

- *Pro*: Restores the cue exactly where a developer would look for it (click a suspect node, see which peers have one-way visibility).
- *Con*: Costs one rendering branch in `_highlightConnections`. Not much.

**(ii) Distinct base-view style for asymmetric pairs.** Render pairs where only one direction exists (or where directions differ dramatically) with a different edge style — e.g. a dashed-half line, a two-tone color split, or a different stroke. Always visible, no click required.

- *Pro*: Asymmetries become scannable at a glance across the whole mesh. Best for the developer audience.
- *Con*: More visual noise in the default view. Already competing with the offline-dashed style (`isToUnknown || hasOfflineEndpoint`). Needs a style that doesn't clash and still reads cleanly on both light and dark themes.

**(iii) Panel-only "reverse only" tag.** When the chosen edge is incoming and no outgoing edge exists, tag the row: "← reverse only" badge with tooltip "Node X reports this peer, but this peer does not report Node X (possible range / power / parent issue)".

- *Pro*: Cheapest. Purely informational, no graph changes.
- *Pro*: Educates users on what the asymmetry implies.
- *Con*: Requires the user to click each suspect node to find asymmetries — can't scan the mesh for them.

**(iv) Combination: (iii) always + (i) on highlight.** Panel row tells you when asymmetry exists; graph shows the arrow when you click. Low cost, covers both usage patterns.

My lean: **(iv)**. (iii) alone is OK for HA power users. (i) alone is OK for developers. Doing both is cheap and covers both audiences without adding default-view noise. Rolling (ii) in later is an option if we find people still miss asymmetries.

## What I'd like from you

- **1.**: A or B? (Or C if you think the two-edge render is worth trying.)
- **2.**: Agreement that it's follow-up, not a blocker?
- **3..**: Which of (i)(ii)(iii)(iv) — or a different take?

Happy to open follow-up PRs for any of these rather than piling more into this one. No rush.

AlixBa:
I'll try to answer

> choice rule differs between graph and panel.

Unless I missed something, it should not. 

The scenario you're describing can happen, indeed. But the panel, if I am right, is only displayed when you highlight a specific node. For consistency reasons, on a specific node, the edge of this node takes over the edge from the remote node (if existing). This is also reflected in the panel and makes the view consistent "from this node". That's why you should see changing colors on some edges when moving from nodes to nodes.

Let's consider 

`A->B: strong`, `B->A: medium` 
(A= 18446744065119617043= ALPSTUGA; B=18446744065119617046=GRILLPLATS)

- No filter, no highlight: `A-B` displayed as `medium`
- No filter, highlight on `A `: `A-B` displayed as `strong` and showing in the panel
- No filter, highlight on `B`: `A-B` displayed as `medium ` and showing in the panel
- Filter `medium`, no highlight: `A-B` displayed as `strong`
- Filter `medium`, highlight on `A`: `A-B` displayed as `strong` and showing in the panel
- Filter `medium`, highlight on `B`: `A-B` displayed as `strong` and showing in the panel as `(reverse)`

<table border="0">
 <tr>
    <td><b style="font-size:30px">A-B</b></td>
    <td><b style="font-size:30px">B-A</b></td>
 </tr>
 <tr>
    <td>Screenshot 2026-04-22 at 19 57 12</td>
    <td>Screenshot 2026-04-22 at 19 56 48</td>
 </tr>
 <tr>
    <td>Screenshot 2026-04-22 at 19 57 51</td>
    <td>Screenshot 2026-04-22 at 19 57 42</td>
 </tr>
</table>

### 1. Graph and panel can display different signal colors for the same connection before you click anything

Given the options, I'd use A as a first step since you usually want to see weak links to strengthen your network. I'd slightly evolve into C by keeping in mind that the overview of a network should be light and navigable. Having all edges by default is way too much (I tried, hence the left "arrow" comment).

**I wanted to try to double-color edges, depending on the signal strength at the end of the edge but viz.js doesn't allow that**. This was an option to have both colors at the same time without the pollution of duplicating arrows.

**As I was screenshotting I saw that using the connection list on the panel to change the highlight colors 2 nodes in blue**

### 2. Swap-on-click has no transition

I didn't think about the user experience on this indeed. I personally stop the refresh during this as it's moving too much so I totally agree with the point! I'll take any valid option on this.

### 3. Arrow removal eliminates the only visible asymmetric-visibility signal

> (iv) Combination: (iii) always + (i) on highlight

That's what I'd do too. I removed the arrow because of the `(reverse)` but I'd be down to get a more visual signal. Doing this feature I discovered the 2-ways links between nodes (I didn't dive into Thread at all, I just have a network at home I want to see better) and it took me a few minutes to realize that some edges were one-sided.

Apollon77:
Thanks @AlixBa — re-checked the code with your screenshots in hand and pushed a fixup (`dfe8d91`). Breakdown:

### Q1 — choice rule differs between graph and panel

You're right, I was wrong. `_highlightConnections` swaps the visible edge to the highlighted node's outgoing perspective using the hidden edge's base-state color, so post-highlight: graph color == panel color. Pre-highlight: no panel is shown. Your table of states nails it end-to-end. **Closing Q1** — current behavior is correct (pre-click shows worst-case dedup so you can scan the mesh for trouble, on-highlight swaps to the selected node's perspective). C (split-color edges) stays parked as future work — noted on the vis.js limitation.

### Bug you spotted — 2 nodes in blue after panel-list click

Root cause: `_highlightConnections` in `thread-graph.ts` updated the newly selected node's icon to the highlighted variant but only `_clearHighlights` restored the previously highlighted node's icon. When a switch didn't pass through `deselectNode` first — exactly the panel-list path (`selectNodes([B])` replaces vis.js selection but fires `selectNode`, not `deselectNode`) — A's icon stayed "selected" and B's icon also went "selected".

Fixed in `dfe8d91`: icon swap extracted to `_setNodeIconHighlight(id, isHighlighted)` and `_highlightConnections` now resets the previous node's icon before setting the new one. Same-node re-select is also now a no-op (was causing a flicker on dense graphs via the full restore+re-highlight cycle).

### Q2 — swap-on-click has no transition

Agreed, follow-up, not a blocker for this PR.

### Q3 — asymmetric-visibility signal — going with (iv)

Since you were on board, implemented both parts on this branch rather than a separate PR:

- **Graph, on highlight:** truly one-way pairs get an arrow in the data direction (`arrows: "to"`). "Truly one-way" is checked against `_edgePairs` pair data, not edge-base-state, so filter-hidden outgoing edges are not misclassified as asymmetric. The arrow sits at the data `to` end which gives the correct semantics in both "remote reports me" (arrow on me) and "I report remote" (arrow on remote) branches without per-branch direction logic.
- **Panel, always:** new `NodeConnection.isReverseOnly` (true when the pair has no matching outgoing edge at all) renders a red "← one-way" badge with a tooltip explaining the typical causes (range, TX power, stale neighbor table). Reverse views caused by filtering keep the existing italic `(reverse)` hint — the two are now distinguishable.

Format / lint / build / 242 tests all green. Please give it a spin on your network when you get a chance — curious whether the arrow is readable in your dense graph or whether it crosses the "too much visual noise" line.

AlixBa:
Dropping a few screenshots

Screenshot 2026-04-24 at 20 36 14
Screenshot 2026-04-24 at 20 37 27
Screenshot 2026-04-24 at 20 37 37
Screenshot 2026-04-24 at 20 37 59
Screenshot 2026-04-24 at 20 40 47

I feel like arrows are not consistent. In this video, the top orange is arrow-ed but not the green one that appears which is one way. Did I misunderstand your proposal?

https://github.com/user-attachments/assets/1e9e3948-8c81-4617-9db3-bf10c5e00131

Apollon77:
I try to get claude check ;-) (i am still not at home ... sooo can not test myself) :-)

Apollon77:
@AlixBa thanks — pulled the video, that nailed it. Pushed `e6ecf9c`.

### What was happening

You weren't misreading the proposal. The arrow fired only when the pair was **data-level** one-way (only one of `edgeAB`/`edgeBA` exists). When a signal-level filter hides the outgoing direction and the incoming one survives, the pair still has both edges in data — so the strict rule said "symmetric, no arrow". The panel marked the entry `(reverse)` italic, but the graph drew a plain edge identical to a healthy bidirectional link. That's what your green-no-arrow case was.

The orange-with-arrow case in the same view was a genuine data-level one-way (BILRESA), so it got the arrow under the strict rule. Same view, two pairs, two different cause categories → inconsistent visual.

### The fix

In `_highlightConnections`, the **remote-reports-highlighted** branch (line 460) reaches `_asymmetricEdgeUpdate` only after the swap attempt already failed — by definition we have no recoverable outgoing direction (truly absent OR filter-hidden). To the eye those are identical, so always draw the arrow there. Cause is disambiguated in the panel: red `one-way` badge for data-level, italic `(reverse)` for filter-induced.

The **highlighted-reports-peer** branch (line 467) keeps the strict data-asymmetric check — that view is the user's own perspective, and a peer-side filter hide shouldn't add noise to it.

Also switched the arrow setter from the `"to"` shorthand to the explicit `{ to: { enabled: true, scaleFactor: 1 } }` object form. Some vis-network builds skip the shorthand on dashed (offline) edges; the object form is reliable.

### What you should see now

In your video's Frame 2 (medium filter ON, ALPSTUGA aqm 17043 selected): the green edge to GRILLPLATS Plug 17045 — the one the panel marks `(reverse)` — now gets an arrow pointing at the selected node, matching the BILRESA orange-with-arrow case from Frame 1.

Format / lint / build / 242 tests green. Give it a spin on your network when you have a moment.

AlixBa:
houuuu, looks good to me. I didn't spot any non-intended behavior on my side 🤩

Apollon77:
Cool, I will also look at it now that I'm back home and will prepare it for next release. thanks!

Apollon77:
PS: @AlixBa please have a look at the changelog ... is that containing all the things relevant for users to know or should we add something more?

Issue #526, opened 2026-04-10
IKEA Battery devices (e.g.MYGGBETT, MYGGSPRAY) not updating state - WITHOUT a Dirigera hub!


kristbaum:
### Version

0.5.15

### Node.js Version

some

### Operating System

Docker

### Issue Description

The exact same issues as described in #187, just without the Dirigera Hub.

So I guess we need to wait for firmware updates for the individual sensors, not just the hub?

### Steps to Reproduce

1. Restart matter-js
2. Open the windows - State updated in HA
3. Close the windows - State updated in HA
4. Wait a few minutes
5. Open the windows - State is NOT updated in HA
6. Ping matter device - OK
7. Reinterview matter device - State is updated in HA
8. Close the window - State is NOT updated in HA
9. Ping matter device - OK
10. Reinterview matter device - State is updated in HA

### Server Logs

```shell
2026-04-10 18:34:38.403 INFO   ClientInteraction    Subscribe » @1:1c•04d0⇵8c78 min: 0 max: 9m 35s attributes: 1 events: 1
2026-04-10 18:34:43.764 INFO   ClientInteraction    Subscription successful « @1:1c•04d0⇵8c78 2↔2 id: 25ca5ec9 interval: 15m timeout: 15m 39s
2026-04-10 18:34:43.899 INFO   Controller~ndHandler Node @1:1c availability changed to true (state: Connected)
2026-04-10 18:34:43.899 INFO   WebSocketC~erHandler [0] Node @1:1c availability changed to true
2026-04-10 18:34:43.900 INFO   WebSocketC~erHandler [1] Node @1:1c availability changed to true
2026-04-10 18:35:22.440 INFO   ClientEventEmitter   Received event switch.initialPress on server-1-fff1.@1:1c.ep1 newPosition: 1
2026-04-10 18:35:22.609 INFO   ClientEventEmitter   Received event switch.shortRelease on server-1-fff1.@1:1c.ep1 previousPosition: 1
2026-04-10 18:35:22.971 INFO   ClientEventEmitter   Received event switch.multiPressComplete on server-1-fff1.@1:1c.ep1 previousPosition: 1 totalNumberOfPressesCounted: 1
2026-04-10 18:36:00.908 INFO   ClientEventEmitter   Received event switch.initialPress on server-1-fff1.@1:1c.ep2 newPosition: 1
[...]

PR #591, opened 2026-04-30
docs: stateful firewall and conntrack requirements for SED reliability

Apollon77:
## Summary

- Adds prescriptive (do / do-not) guidance to `docs/os_requirements.md` covering the network/kernel pitfalls reported across [#526](https://github.com/matter-js/matterjs-server/issues/526) and related discussions.
- Cross-link added from `docs/docker.md` troubleshooting for the common "battery devices drop subscriptions every 15–30 min" symptom.

Topics covered:
- `nf_conntrack_udp_timeout_stream` — default 120 s is shorter than typical SED MaxIntervals (15–30 min) → raise to ≥ 1800 (recommended 3600) on every kernel that filters Matter traffic, with a check/raise/persist snippet.
- "Where to apply": bare-metal host, VM hypervisor (Proxmox per-VM, vSphere DFW), Docker host firewall (firewalld/ufw), HAOS via Supervisor.
- Don't filter Matter UDP by source or destination port (controller binds ephemeral; device side is not pinned to 5540).
- Multi-TBR caveat: do not conntrack-track traffic forwarded between the Thread interface and the LAN.
- VM/container: NIC directly on the TBR network, IPv6 sysctls inside the guest, `--network=host` for Docker, no mDNS reflectors between segments.
- NetworkManager / systemd-networkd override `sysctl -w` — values must live in the network manager's config.

Docs-only change. No code paths touched.


docs/os_requirements.md
```diff
@@ -73,9 +73,59 @@ sysctl -w net.ipv6.conf.wlan0.accept_ra=1
...
+
 If your system has IPv6 forwarding enabled (not recommended, see above), you'll
 have to use `2` for the accept_ra variable. See also the [Thread Border Router - Bidirectional IPv6 Connectivity and DNS-Based Service Discovery codelab](https://openthread.io/codelabs/openthread-border-router#6).
 
+### Multiple Thread Border Routers (multi-TBR)
+
+**DO NOT** apply conntrack-based firewall rules to traffic **forwarded** between the Thread interface and the LAN on a Thread Border Router itself, when multiple TBRs share the same Thread network. A single Matter flow can ingress one TBR and egress another, so each TBR's conntrack only ever sees half the flow and will drop it.
+
+This applies only to the TBR's forwarding/router role. It does **not** affect conntrack on the Matter Server's own host (see [Stateful firewalls](#stateful-firewalls-host-vm-hypervisor) below) — there, both directions terminate at the same kernel and conntrack works correctly.
```
agners: Ah yeah make sense:

```suggestion
This applies only to the TBR's forwarding/router role. It does **not** affect conntrack on the Matter Server's own host (see [Stateful firewalls](#stateful-firewalls-host-vm-hypervisor) below) — there, traffic of both directions are observed by the same firewall which makes conntrack work correctly.
```

docs/os_requirements.md
```diff
@@ -73,9 +73,59 @@ sysctl -w net.ipv6.conf.wlan0.accept_ra=1
...
+**DO** apply this on the kernel that actually filters the packets:
+
+- Bare-metal: the host itself.
+- VM with hypervisor firewall (e.g. Proxmox per-VM firewall, vSphere DFW): the **hypervisor** kernel. Setting it inside the guest has no effect.
+- Docker on Linux with firewalld (Fedora/RHEL default) or ufw (Ubuntu, when enabled): the Docker **host**.
+- Home Assistant OS via the Supervisor: usually fine out of the box. Only worry if you have added a custom firewall layer.
+
+**DO NOT** filter Matter UDP traffic by destination port. The controller binds an ephemeral port, and device source ports also vary.
+
+**DO NOT** pin firewall rules to UDP/5540. Most devices use it, but not all.
```
agners: In my mind, the Controller is typically the one initiating initial traffic, so if we setup a stateful firewall, then we'd do this based on traffic originating from the Matter server. Now that the originating port is ephemeral make sense, and that is the one connection tracking needs to remember.

But isn't the destination port spec mandated? Or can it be different depending on mDNS?

I feel if we can't rely on source and destination port, then there is not much we can go after to even setup a firewall 😅

agners (commented):
I think this information are quite helpful, especially the connection tracking.

But I'd somewhere conclude or hint that whenever problems appear, users should try with disabled firweall before making issues. Maybe we can also add that to the issue tracker template.

docs/os_requirements.md
Apollon77: In fact 5540 is "just" the default assigned port for matter, but basically the  port to use is variable and can change and all that matters is that MDNS announces are correct. So basically nothing we can rely on for any side. Sure 99,x% will be 5540 but not 100%

agners: Right ok. Maybe we can add the information that while 5540 is the IANA assigned port, devices are free to choose a different port which they will announce in their mDNS announcements.

```suggestion
**DO NOT** filter Matter UDP traffic by destination port. The controller binds an ephemeral port, and device source ports also vary.

**DO NOT** pin firewall rules to UDP/5540. While 5540 is the IANA-assigned Matter port and is commonly used, the Matter specification does not require devices to use a single predetermined well-known port. A Matter node may listen on a different UDP port and advertise that port through its mDNS/DNS-SD service records. Controllers are expected to use the advertised port rather than assuming 5540.
```

docs/os_requirements.md
```diff
@@ -73,9 +73,60 @@ sysctl -w net.ipv6.conf.wlan0.accept_ra=1
...
+echo 'net.netfilter.nf_conntrack_udp_timeout_stream = 3600' | sudo tee /etc/sysctl.d/99-matter.conf
+# Present and ≥ 1800 → done.
+```
+
+**DO** apply this on the kernel that actually filters the packets:
+
+- Bare-metal: the host itself.
+- VM with hypervisor firewall (e.g. Proxmox per-VM firewall, vSphere DFW): the **hypervisor** kernel. Setting it inside the guest has no effect.
+- Docker on Linux with firewalld (Fedora/RHEL default) or ufw (Ubuntu, when enabled): the Docker **host**.
+- Home Assistant OS via the Supervisor: usually fine out of the box. Only worry if you have added a custom firewall layer.
```
agners:
```suggestion
- Home Assistant OS: Configured correctly out of the box. Only worry if you have added a custom firewall layer (e.g. VM with hypervisor firewall).
```

agners (approved):

docs/os_requirements.md
```diff
@@ -73,9 +73,60 @@ sysctl -w net.ipv6.conf.wlan0.accept_ra=1
...
+- Bare-metal: the host itself.
+- VM with hypervisor firewall (e.g. Proxmox per-VM firewall, vSphere DFW): the **hypervisor** kernel. Setting it inside the guest has no effect.
+- Docker on Linux with firewalld (Fedora/RHEL default) or ufw (Ubuntu, when enabled): the Docker **host**.
+- Home Assistant OS: Configured correctly out of the box. Only worry if you have added a custom firewall layer (e.g. VM with hypervisor firewall).
+
+**DO NOT** filter Matter UDP traffic by destination port. The controller binds an ephemeral port, and device source ports also vary.
+
+**DO NOT** pin firewall rules to UDP/5540. While 5540 is the IANA-assigned Matter port and is commonly used, the Matter specification does not require devices to use a single predetermined well-known port. A Matter node may listen on a different UDP port and advertise that port through its mDNS/DNS-SD service records. Controllers are expected to use the advertised port rather than assuming 5540.
+
+If you must filter, scope rules by source IPv6 prefix (Thread ULA + OMR prefix) or by interface — never by port.
```
M0rious: If you must filter, scope rules by network interface or by source IPv6 prefix (Thread ULA + OMR prefix) — never by port. On OTBR, `ot-ctl prefix` lists the OMR prefix (look for the `paros` flags).

docs/os_requirements.md
```diff
@@ -73,9 +73,60 @@ sysctl -w net.ipv6.conf.wlan0.accept_ra=1
...
+
+Diagnostic sentinel: a per-instance firewall chain drop counter climbing in step with battery-device report cycles while your "Matter" ACCEPT rule shows zero hits → conntrack timeout is the cause.
```
M0rious: Diagnostic sentinel: a per-instance firewall chain drop counter climbing in step with battery-device report cycles while your "Matter" ACCEPT rule shows zero hits → conntrack timeout is the cause. To confirm, run `conntrack -L | grep <controller-ip>` on the filtering kernel — each Matter flow shows its remaining UDP timeout in seconds. The `[ASSURED]` flag plus the countdown is the unambiguous tell.

PR #599, opened 2026-05-03
Standardize MatterClient access via Lit context (#421)

### LINKED ISSUE #421: Standardize dashboard client access pattern
## Problem

Dashboard components use three different patterns to access `MatterClient`:
1. Required property: `@property() public client!: MatterClient` (most dialogs)
2. Optional property: `public client?: MatterClient` (server-details, header)
3. Lit context: `@consume({ context: clientContext })` (commission-node children)

This inconsistency makes refactoring risky and creates nullable confusion.

## Proposal

Standardize on Lit context (`@consume`) for all components that need the client. This is the most Lit-idiomatic approach and eliminates manual property passing through show functions.

markvp:
## Summary

- Introduces a `ContextProvider` in `matter-dashboard-app` that broadcasts the `MatterClient` instance to all descendant components via `@lit/context`
- Converts all 17 dashboard components and dialogs from manual `.client` prop drilling to `@consume({ context: clientContext, subscribe: true })` decorators
- Removes the `client` argument from `showCommissionNodeDialog()`, `showLogLevelDialog()`, and `showNodeBindingDialog()` helpers — dialogs now receive the client from context automatically

## Test plan

Manual validation performed against a live server instance (`npm run server -- --storage-path data`) with a diagnostics dump imported as test nodes:

- [x] Dashboard loads and connects at `http://localhost:5580` — server overview renders correctly
- [x] Settings cog in header opens log level dialog (tests `dashboard-header` → `showLogLevelDialog` context path)
- [x] Diagnostics dump import via file upload prompt works (`server-details.importTestNode`)
- [x] Test node appears in node list after import
- [x] Node view (`#node/<id>`) renders node details correctly (`matter-node-view`, `node-details`)
- [x] Endpoint view (`#node/<id>/<endpoint>`) renders endpoint and cluster list (`matter-endpoint-view`, `node-details`)
- [x] Cluster view (`#node/<id>/<endpoint>/<cluster>`) renders attributes and command panel (`matter-cluster-view`, `base-cluster-commands`)
- [x] Commission node dialog opens from server overview (`showCommissionNodeDialog`)
- [x] Back navigation works correctly at each level
- [x] No JavaScript console errors throughout

Build and lint:
- [x] `npm run format` — passes
- [x] `npm run lint` — 0 errors, 0 warnings
- [x] `npm run build` — passes

Closes #421 


packages/dashboard/src/pages/matter-dashboard-app.ts
```diff
@@ -198,7 +198,6 @@ class MatterDashboardApp extends LitElement {
             // cluster level
             return html`
                 <matter-cluster-view
-                    .client=${this.client}
                     .node=${this.client.nodes[this._route.path[0]]}
                     .endpoint=${parseInt(this._route.path[1], 10)}
                     .cluster=${parseInt(this._route.path[2], 10)}
```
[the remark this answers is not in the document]
markvp: Fixed in 264b4c9 — added `this.provider.setValue(clone(this.client))` immediately after `startListening()` resolves, so all consumers receive a live client value before any `nodes_changed` or `server_info_updated` event fires.

packages/dashboard/src/pages/network/network-details.ts
```diff
@@ -704,7 +704,6 @@ export class NetworkDetails extends LitElement {
                 ${this._showUpdateDialog
                     ? html`
                           <update-connections-dialog
```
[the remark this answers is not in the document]
markvp: Fixed in 264b4c9 — added `subscribe: true` to the `@consume` decorator, consistent with every other `clientContext` consumer in the codebase.

markvp:
fixes #421

Apollon77 (commented):
## Review summary

Great direction — context replaces the prop-drilling chain cleanly. Build/lint pass. Two correctness concerns and several consistency follow-ups before merge.

### 🔴 Critical
- **Clone-induced `msgId` divergence → silent command collisions** (see inline on `matter-dashboard-app.ts:111`)

### 🟡 Important
- Settings dialog chain not migrated — partial standardization (inline on `header.ts`)
- Leftover `.client=` bindings on now-consuming children (inline on cluster/endpoint/node-view + network-details)
- `network-details.ts` `subscribe: false → true` not justified by code reads (inline)
- Type lies: `client?: MatterClient` → `client!: MatterClient` paired with runtime `if (this.client)` guards (inline on header + server-details)

### ⚪ Nits (no in-diff anchor)

**1. `ContextProvider(this, { context: clientContext, initialValue: this.client })` (`matter-dashboard-app.ts:51`)** — field initializer runs during element construction; `dashboard.client = client` in `entrypoint/main.ts` runs after construction. So `initialValue` is always `undefined`. Reads as if seeding a value; doesn't. Drop `initialValue` or move provider creation into a setter/`firstUpdated` after assignment.

**2. `@consume({subscribe:true}) @property({attribute:false})` stacking** — `@property` is technically redundant; `@consume` already triggers `requestUpdate` on subscribed value changes. Harmless but noise. Inconsistent across files (some have it, `matter-cluster-view.ts` etc. didn't originally). Pick one and unify.

**3. `packages/dashboard/src/util/clone_class.ts` (not in diff but now load-bearing for context propagation):**
   - Uses `any` and lacks return type — violates CLAUDE.md "avoid type casts / use generics":
     ```ts
     export const clone = <T extends object>(orig: T): T =>
         Object.assign(Object.create(Object.getPrototypeOf(orig)), orig);
     ```
   - Shallow `Object.assign` skips ES private fields (`#x`), accessor-backed state, and non-enumerable own properties. If `MatterClient` ever grows a `#field`, clone silently loses it.
   - If you fix the critical msgId concern by removing the clone-on-publish pattern (recommended), this file may become unused — preferable.

The msgId issue is the only blocker — others are polish to deliver the "standardize" promise the title makes.

packages/dashboard/src/pages/matter-dashboard-app.ts
```diff
@@ -108,6 +108,7 @@ class MatterDashboardApp extends LitElement {
         this.client.startListening().then(
             () => {
                 this._state = "connected";
+                this.provider.setValue(clone(this.client));
```
Apollon77: 🔴 **Critical: `clone(this.client)` causes `msgId` divergence → silent command collisions.**

`MatterClient.msgId` is a TS-private *primitive* (`packages/ws-client/src/client.ts:55`) incremented by `sendCommand` (`client.ts:371`). `result_futures` (`client.ts:46`) is a shared object reference. The shallow `Object.assign(Object.create(proto), orig)` in `util/clone_class.ts` copies `msgId` by value but shares `result_futures` by reference.

Failure path:
1. Original sends `startListening` → `original.msgId = V`.
2. `setValue(clone-A)` published; `clone-A.msgId = V`. UI sends command → `clone-A.msgId = V+1`, writes `result_futures["V+1"] = {clone-A's resolve}`.
3. `nodes_changed` fires (frequent — every attribute storm) → `setValue(clone-B)` made from `original` whose `msgId` is still `V` → `clone-B.msgId = V`.
4. UI sends another command via `clone-B` → `clone-B.msgId = V+1`, writes `result_futures["V+1"] = {clone-B's resolve}`, **overwriting clone-A's pending entry**.
5. Server reply for the first command resolves clone-B's promise. Clone-A's promise hangs until `commandTimeout` (default 5 minutes) and rejects with `CommandTimeoutError`.

Silent, intermittent, scales with UI activity during attribute updates.

**Fix**: don't snapshot the client. Keep the context value as the live `MatterClient` ref and trigger consumer re-renders via a separate signal — e.g. a `tickContext` that increments, a custom event on the host, or a Lit reactive controller. Same applies to the two `setValue(clone(...))` calls in `_setupEventListeners`.

packages/dashboard/src/pages/components/header.ts
```diff
@@ -35,7 +37,9 @@ export class DashboardHeader extends LitElement {
     @property({ type: Boolean }) public hasThreadDevices?: boolean;
     @property({ type: Boolean }) public hasWifiDevices?: boolean;
 
-    public client?: MatterClient;
+    @consume({ context: clientContext, subscribe: true })
+    @property({ attribute: false })
+    public client!: MatterClient;
```
Apollon77: 🟡 **Inconsistent migration — settings dialog chain still prop-drills.**

`_openSettings` (line 74) still calls `showSettingsDialog(this.client)`. Inside that chain:
- `show-settings-dialog.ts:9-12` — receives `client`, assigns `dialog.client = client`
- `settings-dialog.ts:23` — bare `@property` (no `@consume`)
- `settings-dialog.ts:92` — passes `.client=${this.client}` to `<log-level-section>`
- `log-level-section.ts` — bare `@property`

Sibling dialogs (commission/log-level/binding) were migrated. Either finish the settings chain for consistency or scope-clarify in the PR description.

Also a type-lie nit on this line: `client!: MatterClient` (non-null assertion) but lines 73, 165, 178, 190 all guard with `if (this.client)` — runtime admits `undefined`. Original `client?: MatterClient` was more honest. Either keep `?:` or remove the guards.

packages/dashboard/src/pages/components/server-details.ts
```diff
@@ -11,18 +11,22 @@ import "@material/web/divider/divider";
...
 import { showCommissionNodeDialog } from "../../components/dialogs/commission-node-dialog/show-commission-node-dialog.js";
 import "../../components/ha-svg-icon";
 import { handleAsync } from "../../util/async-handler.js";
 
 @customElement("server-details")
 export class ServerDetails extends LitElement {
-    public client?: MatterClient;
+    @consume({ context: clientContext, subscribe: true })
+    @property({ attribute: false })
+    public client!: MatterClient;
```
Apollon77: Type lie: `client!: MatterClient` paired with `if (!this.client) return html\`\`;` on line 32. The runtime guard says it can be `undefined`; the `!` says it can't. Original `client?: MatterClient` was correct. Pick a stance — the codebase is inconsistent (this file asserts, `header.ts` guards).

packages/dashboard/src/pages/matter-cluster-view.ts
```diff
@@ -130,12 +133,11 @@ class MatterClusterView extends LitElement {
             <dashboard-header
                 .title=${`Node ${this.node.node_id} ${nodeHex}  |  Endpoint ${this.endpoint}  |  Cluster ${this.cluster} (${clusterName})`}
```
Apollon77, line 136: Leftover `.client=` binding to flag — line 114 (the "Not found" branch) still passes `.client=${this.client}` to `<dashboard-header>`, even though `<dashboard-header>` now consumes from context. The success branch was cleaned up here; the not-found branch was missed. Drop for consistency.

packages/dashboard/src/pages/matter-endpoint-view.ts
```diff
@@ -79,12 +83,11 @@ class MatterEndpointView extends LitElement {
             <dashboard-header
                 .title=${`Node ${this.node.node_id} ${nodeHex}  |  Endpoint ${this.endpoint}`}
                 .backButton=${`#node/${this.node.node_id}`}
-                .client=${this.client}
             ></dashboard-header>
```
Apollon77, line 91: Leftover `.client=` binding — line 66 ("Not found" branch) still passes `.client=${this.client}` to `<dashboard-header>`. Success branch cleaned, not-found missed. Drop.

packages/dashboard/src/pages/matter-node-view.ts
```diff
@@ -70,11 +74,7 @@ class MatterNodeView extends LitElement {
         const nodeHex = formatNodeAddress(fabricIndex, this.node.node_id);
 
         return html`
-            <dashboard-header
-                .title=${`Node ${this.node.node_id} ${nodeHex}`}
-                .client=${this.client}
-                backButton="#"
-            ></dashboard-header>
+            <dashboard-header .title=${`Node ${this.node.node_id} ${nodeHex}`} backButton="#"></dashboard-header>
```
Apollon77, line 79: Leftover `.client=` binding — line 53 ("Node not found" branch) still passes `.client=${this.client}` to `<dashboard-header>`. Drop for consistency with the success branch.

packages/dashboard/src/pages/network/network-details.ts
```diff
@@ -71,7 +71,7 @@ export class NetworkDetails extends LitElement {
     @property({ type: Object })
     public threadEdgePairs: Map<string, ThreadEdgePair> = new Map();
 
-    @consume({ context: clientContext })
+    @consume({ context: clientContext, subscribe: true })
```
Apollon77: 🟡 `subscribe: false → true` is not justified by code reads in this file. Only client access is `this.client?.serverInfo?.fabric_index` (line 124). `serverInfo` is a getter that reads through to the shared `connection.serverInfo` object, mutated in-place on `server_info_updated`. All other reactive data flows in via props (`nodes`, `borderRouters`, etc.). Subscribing here only adds redundant re-renders on every `nodes_changed` (frequent). Revert to `subscribe: false` unless there's a specific case I'm missing.

Also: lines 1084 and ~1162 — there are still leftover `.client=${this.client}` bindings on `<update-connections-dialog>` in some branches even though that child now consumes from context. Drop them.

packages/dashboard/src/pages/matter-dashboard-app.ts
markvp: Fixed — and updated further after reflection.

Initial fix replaced `clone(this.client)` with `provider.setValue(this.client, true)`, which solved the identity problem but used `force: true` as an escape hatch rather than a proper design.

On review, your suggested approach is the right one. Switched to a dedicated `tickContext`:

- `clientContext` is set **once** after `startListening()` resolves — a stable reference that never changes, so no `force` or subscription needed.
- `tickContext` is a plain incrementing number, published on every `nodes_changed` / `server_info_updated` event. Consumers subscribe to this for re-renders.
- The two concerns are now explicitly separated: stable ref for data, tick for change signal.

`clone_class.ts` is deleted. `settings-dialog` and `log-level-dialog` drop their client consume entirely — neither uses `this.client` directly after `log-level-section` migrated to context.

packages/dashboard/src/pages/components/header.ts
markvp: Both addressed.

**Settings chain**: `showSettingsDialog` now takes no arguments. `settings-dialog` and `log-level-section` both use `@consume({ context: clientContext, subscribe: true })`. The dialog is appended to `matter-dashboard-app`'s `renderRoot`, so it sits inside the provider's shadow tree and receives context naturally — consistent with how commission/log-level/binding dialogs were already handled.

**Type lie**: `client!: MatterClient` → `client?: MatterClient`, `@property({ attribute: false })` removed (see nit below on `@consume + @property` stacking).

packages/dashboard/src/pages/components/server-details.ts
markvp: Fixed: `client!: MatterClient` → `client?: MatterClient`, `@property({ attribute: false })` removed. The one access site not covered by the existing `if (!this.client) return html``` guard — the `importTestNode` call inside a `reader.onload` closure — was updated to use optional chaining.

packages/dashboard/src/pages/matter-cluster-view.ts
markvp: Removed.

packages/dashboard/src/pages/matter-endpoint-view.ts
markvp: Removed.

packages/dashboard/src/pages/matter-node-view.ts
markvp: Removed.

packages/dashboard/src/pages/network/network-details.ts
markvp: Both addressed.

**subscribe**: reverted to `subscribe: false` (the default, so the option is just dropped). You're right — `serverInfo` is a getter over a mutated-in-place object and all other reactive data flows in via props; no subscription is warranted here.

**Leftover `.client=`**: the binding at line 1084 on `<update-connections-dialog>` in the border-router branch is removed. The instance at line ~1161 (node branch) never had it — that one was already clean.

markvp:
Addressing the review body nits:

**1. `initialValue: this.client`**
Dropped. The field initialiser runs during element construction before `dashboard.client = client` is assigned, so the value was always `undefined`. The provider is now constructed with just `{ context: clientContext }`.

**2. `@consume + @property` stacking**
Removed the redundant `@property({ attribute: false })` from every `@consume`-decorated `client` field across all consumer files. `@consume({ subscribe: true })` already calls `requestUpdate()` on value changes; the `@property` was pure noise. Unified on `@consume`-only throughout.

**3. `clone_class.ts`**
Deleted — unused once the clone-on-publish pattern was removed.

Apollon77:
@markvp Could it be that there was a code push missing? I see no code changes but your comments state you changed things?!

markvp:
> @markvp Could it be that there was a code push missing? I see no code changes but your comments state you changed things?!

@Apollon77 eeek, correct! rebased and pushed now. sorry! 🤦‍♂️

Apollon77:
Thanks

PR #600, opened 2026-05-03
Migrate commissioning methods to new Node API (#428)

### LINKED ISSUE #428: Migrate commissioning methods to new Node API
Several methods in ControllerCommandHandler.ts still use the legacy commissioning API. These should be migrated to use the new Node API when it stabilizes.

Affected locations:
- `ControllerCommandHandler.ts:1249` - `commissionOnNetwork` method
- `ControllerCommandHandler.ts:1300` - `removeFabric` method

markvp:
## Summary

- Replaces `commandsOf(OperationalCredentialsClient)` in `removeFabric` with `#invokeCommand` using `OperationalCredentials.Cluster` — removes the last use of the legacy behavior-client pattern in this file
- Extracts ACL and binding conversion logic from `setAclEntry` and `setNodeBinding` into exported pure functions (`convertAclEntries`, `convertBindingTargets`) so they are independently testable
- Adds `ControllerCommandHandlerTest.ts` with 15 unit tests covering both helpers: field mapping, null/undefined propagation, bigint handling, `FabricIndex.OMIT_FABRIC`, and multiple-entry cases

Closes #428

## Test plan

- [ ] `npm test` — all unit tests pass (71/71 in ws-controller, pre-existing integration ping failure unchanged)
- [ ] `npm run build` — clean build
- [ ] `npm run lint` — no errors

### DIFF
--- a/packages/ws-controller/src/controller/ControllerCommandHandler.ts
+++ b/packages/ws-controller/src/controller/ControllerCommandHandler.ts
@@ -27,7 +27,6 @@ import {
     Timer,
     DnsRecordType,
 } from "@matter/main";
-import { OperationalCredentialsClient } from "@matter/main/behaviors";
 import {
     AccessControl,
     BasicInformation,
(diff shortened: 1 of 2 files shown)


Apollon77:
Sorry but this is exactly not what the topic was:
1.) Replacement for "removeFabric" replaced it from a Fully types and "already partly new Node API" to an untyped generic invoke .. so this is not whats the task is about. The task is about to move "removeFabric" to decommission on the ComissionignClient ... and other cases ... let me do that when it is worth it
2.) Exposeing simply conversion tools into own publish methods just for testability also has nothing to do with the task

I close the PR.

PR #630, opened 2026-05-09
custom-clusters: add TCL vendor cluster (VID 0x1334)

iamadamreed:
## What
Adds the TCL vendor-specific Matter cluster used by TCL dehumidifiers.

| Cluster | ID | Purpose |
|---|---|---|
| `TclDehumidifierCluster` | `0x1334FC03` | `mode` (enum8 — Set/Continue/Comfort/Smart/Dry), `targetHumidity`, `currentHumidity`, `waterBucketFull`, `filterAlert`, `errorCodes`, `featureSet` |

## Why
TCL CSA-registers their Matter dehumidifiers (e.g. **H50D44W**, **H50D66KW**) as a generic Fan device type (`0x002B`). Without a registered TLV schema for the vendor-specific cluster, `write_attribute` on attributes like target humidity silently no-ops, so users cannot control the dehumidifier from a generic Matter controller (Apple Home, Google Home, Home Assistant matter integration).

Empirically verified against an H50D44W running firmware 1.0 paired to matter-server in matter.js mode (`beta=true`). Live attribute reads on cluster `0x1334FC03`:

```
attr 0 (mode):            enum8, writable, current=0  → Set
attr 1 (targetHumidity):  uint8, writable, current=50
attr 2 (currentHumidity): uint8,           current=54
attr 3 (waterBucketFull): bool,            current=false
attr 4 (filterAlert):     bool,            current=false
attr 5 (errorCodes):      string,          current="[]"     (literal JSON-encoded text — not a Matter list TLV)
attr 6 (featureSet):      string,          current="[3]"    (same — JSON-encoded text)
```

A full diagnostic dump (BasicInformation, Descriptor on endpoint 1, `AttributeList` confirming the cluster definition matches the device 1:1) is in the [first issue comment](https://github.com/matter-js/matterjs-server/pull/630#issuecomment-4416017065).

## Pattern
Mirrors `heiman.ts` (most recent vendor add). All checks pass:
- `npm run format` ✓
- `npm run lint` ✓ (0 warnings, 0 errors)
- `npm run build` ✓ (custom-clusters package + dashboard descriptions auto-regenerate)
- `npm run python:generate` ✓ (auto-generated `chip/clusters/cluster_defs/TclDehumidifierCluster.py` is included in the diff)

## Test plan
- [x] Type-check + lint clean
- [x] `descriptions.ts` regenerated includes the cluster at ID 322239491
- [x] Auto-generated Python client class (`TclDehumidifierCluster.py`) committed alongside the TS source
- [x] End-to-end verified on a real H50D44W: `write_attribute` lands on the device, value persists across re-reads
- [ ] Optional follow-up: Home Assistant core matter integration discovery schemas (separate PR; unblocks proper humidifier/sensor/binary_sensor/select entities for TCL devices out of the box)

## Mode mapping
`mode` is `enum8` with the following values, captured as a `const enum TclMode` alongside the cluster (the wire encoding stays an 8-bit value and the symbolic mapping is documented in source):

| Value | Name |
|---:|---|
| 0 | `Set` |
| 1 | `Continue` |
| 2 | `Comfort` |
| 3 | `Smart` |
| 4 | `Dry` |

Order matches what's shown in TCL's mobile app.

## Review history

Earlier revisions of this PR also defined a `TclPrivateCluster` (`0x1334FC00`) with a single opaque attribute at vendor-prefixed ID `0x1334E000`. Per maintainer review, that cluster was dropped — its contents are unknown and declaring it added noise without signal. Devices still expose the cluster in their `Descriptor.ServerList`; matter.js's default handling of undefined clusters is sufficient. If the vendor-prefixed attribute's contents are ever identified, it can return as a follow-up PR with the right type.

## Consumers / downstream

- **[iamadamreed/ha-tcl-matter](https://github.com/iamadamreed/ha-tcl-matter)** — Home Assistant custom integration that consumes this cluster and surfaces `humidifier.tcl_dehumidifier`, `binary_sensor.tcl_dehumidifier_water_bucket_full`, `select.tcl_dehumidifier_mode`, etc. Released as v0.4.x with live reads + anti-loop write semantics. End-to-end verified.
- **[iamadamreed/addons](https://github.com/iamadamreed/addons)** — fork of `home-assistant/addons` that ships a patched Matter Server addon (`8.4.0-tclpatch.X`) bundling this PR's cluster decoder so users can run the integration today while this PR awaits review. Retires once this PR merges and a stock matter-server release ships.

The full chain is documented in [the integration's PROJECT_STATUS.md](https://github.com/iamadamreed/ha-tcl-matter/blob/main/PROJECT_STATUS.md).


packages/custom-clusters/src/clusters/tcl.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    /** True when the water bucket needs to be emptied. */
+    @attribute(0x0003, bool)
+    waterBucketFull?: boolean;
+
+    /** Filter / child-lock alert (semantics device-specific). */
+    @attribute(0x0004, bool)
+    filterAlert?: boolean;
+
+    /** Active error codes as a JSON-encoded list, e.g. "[]" or "[3]". */
```
Apollon77: really a json encoded string? Or a real array of numbers?
ideally show the diagnostics of such a device then we can verify this

packages/custom-clusters/src/clusters/tcl.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * Mapping verified empirically against an H50D44W on firmware 1.0:
+ * - Mode integer values are observed; the symbolic mapping (Set / Continue /
+ *   Comfort / Smart / Dry) follows the order shown in the TCL Home app and
+ *   is the consumer's responsibility to translate.
+ */
+@cluster(0x1334fc03)
+export class TclDehumidifierCluster {
+    /** Operating mode — 0=Set, 1=Continue, 2=Comfort, 3=Smart, 4=Dry. */
+    @attribute(0x0000, uint8, writable)
```
Apollon77: Should we define that as real enum?

packages/custom-clusters/src/clusters/tcl.ts
```diff
@@ -0,0 +1,68 @@
...
+    @attribute(0x0003, bool)
+    waterBucketFull?: boolean;
+
+    /** Filter / child-lock alert (semantics device-specific). */
+    @attribute(0x0004, bool)
+    filterAlert?: boolean;
+
+    /** Active error codes as a JSON-encoded list, e.g. "[]" or "[3]". */
+    @attribute(0x0005, string)
+    errorCodes?: string;
```
Apollon77, line 75: string?

packages/custom-clusters/src/clusters/tcl.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    /** Supported feature set as a JSON-encoded list, e.g. "[3]". */
+    @attribute(0x0006, string)
+    featureSet?: string;
```
Apollon77, line 83: same here, string or array?

packages/custom-clusters/src/clusters/tcl.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * TCL private/identification cluster — Vendor ID 0x1334.
+ *
+ * Holds an opaque vendor-prefixed string attribute that is populated by
+ * the TCL Home app at first commissioning. Empty on devices commissioned
+ * directly without going through TCL Home.
+ */
+@cluster(0x1334fc00)
+export class TclPrivateCluster {
```
Apollon77: Ok if the content is unknown (and really a string?) Do we need to define it here when we have no clue what it means?

Apollon77 (commented):
Please provide a diagnostic dump of such a device that I can verify the datatypes and details

iamadamreed:
Diagnostic dump from a paired **TCL H50D44W** dehumidifier on firmware **1.0**, paired to `python-matter-server` (8.4.0 + this PR's cluster decoder) over Apple Home + Home Assistant fabrics. Captured via Home Assistant's built-in `matter` integration diagnostics export (which dumps `MatterClient.nodes[*].attributes`).

### BasicInformation (endpoint 0, cluster `0x0028`)

| Attr | Name | Type | Value |
|---:|---|---|---|
| 0 | `DataModelRevision` | uint16 | `17` |
| 1 | `VendorName` | string | `"TCL"` |
| 2 | `VendorID` | uint16 | `4916` (`0x1334`) |
| 3 | `ProductName` | string | `"TCL Dehumidifier"` |
| 4 | `ProductID` | uint16 | `32770` (`0x8002`) |
| 5 | `NodeLabel` | string | `""` |
| 7 | `HardwareVersion` | uint16 | `1` |
| 8 | `HardwareVersionString` | string | `"1"` |
| 9 | `SoftwareVersion` | uint32 | `1` |
| 10 | `SoftwareVersionString` | string | `"1.0"` |
| 18 | `SerialNumber` | string | `"45EDA565······"` (redacted) |
| 19 | `CapabilityMinima` | struct | `{caseSessionsPerFabric: 3, subscriptionsPerFabric: 3}` |

### Descriptor on endpoint 1 (the device itself, cluster `0x001D`)

```
DeviceTypeList (attr 0): [{ deviceType: 0x002B (Fan), revision: 3 }]
ServerList     (attr 1): [
  0x001D Descriptor,
  0x0003 Identify,
  0x0004 Groups,
  0x0202 FanControl,
  0x0006 OnOff,
  0x1334FC00 (vendor cluster — contents unknown; not added by this PR),
  0x1334FC03 TclDehumidifierCluster,
]
```

This confirms the rationale in the PR description — TCL CSA-registered the device as a generic **Fan** (`0x002B`), so all dehumidifier control lives in the vendor cluster.

### TclDehumidifierCluster `0x1334FC03` (322239491) on endpoint 1

| Attr | Name | Type | Live value | Verified |
|---:|---|---|---|---|
| 0 | `mode` | enum8 (`TclMode`) | `0` → `Set` | mode 0 confirmed against TCL Home app display |
| 1 | `targetHumidity` | uint8 | `50` | writable, persists across re-reads |
| 2 | `currentHumidity` | uint8 | `54` | onboard RH sensor |
| 3 | `waterBucketFull` | bool | `False` | correlates with bucket sensor |
| 4 | `filterAlert` | bool | `False` | filter alert; reset by app |
| 5 | `errorCodes` | string | `"[]"` | literal JSON-encoded text on the wire |
| 6 | `featureSet` | string | `"[3]"` | static; same JSON-encoded-text wire format |
| 65531 | `AttributeList` | list[uint32] | `[0..6, 65528..65533]` | matches PR cluster def 1:1 |
| 65533 | `ClusterRevision` | uint16 | `1` | |

Notes on data types:
- **`mode`** is now `enum8` with a `const enum TclMode { Set=0, Continue=1, Comfort=2, Smart=3, Dry=4 }` declared alongside the cluster. Wire encoding stays an 8-bit value; the symbolic mapping is captured in source. Order verified against the TCL Home app.
- `targetHumidity` (1) and `currentHumidity` (2) are `uint8`. Bucket and filter (3, 4) are `bool` and toggle in real time when the bucket is removed / filter is reset.
- **`errorCodes` (5) and `featureSet` (6) are JSON-encoded strings, not Matter list TLVs.** TCL's firmware emits literal text payloads (`"[]"`, `"[3]"`) — modeling them as `array[uint16]` would tell matter.js to expect a list-TLV header that doesn't exist on the wire and would break reads. Treating them as `string` matches what the device actually sends; consumers JSON-parse application-side if they want a numeric list.
- All seven data attributes plus standard metadata (65528, 65529, 65531, 65532, 65533) are listed in `AttributeList`, matching the PR's attribute map exactly.

### `0x1334FC00` (TCL private vendor cluster) — not added by this PR

The device's `Descriptor.ServerList` also includes cluster `0x1334FC00`, with a single attribute at vendor-prefixed ID `0x1334E000` (currently empty string on devices commissioned outside the TCL Home app). An earlier revision of this PR defined this as `TclPrivateCluster` with the opaque attribute typed as `octstr`, but per maintainer feedback the cluster has been **dropped** — declaring a cluster whose contents we can't interpret adds noise without signal. matter.js's default handling of undefined clusters is sufficient. If/when the contents are identified, the cluster can return as a follow-up PR.

### Write verification

With the cluster decoder this PR adds, `write_attribute` lands on the device end-to-end:
- `1/322239491/1 = 47` → device persists, re-read returns `47`
- `1/322239491/0 = 2` → unit display flips to **Comfort** (`TclMode.Comfort`)

Without the PR's TLV schema, matter.js silently skips the wire encode (the controller acks the write but the value never reaches the device) — that's the symptom that motivated this PR.

---

*Comment updated 2026-05-10 to reflect the tightened cluster definition (`d7a3632`) — `mode` is now `enum8` with a named-values `TclMode`; the speculative private cluster has been dropped per maintainer review.*

packages/custom-clusters/src/clusters/tcl.ts
iamadamreed: Done in `d7a3632` — the attribute is now `enum8` with a `const enum TclMode { Set=0, Continue=1, Comfort=2, Smart=3, Dry=4 }` declared alongside the cluster. Generated Python client surfaces this as `Optional[uint]` (no regression vs. the previous `uint8`); dashboard `descriptions.ts` renders it as `Optional[unknown]`, which matches every other custom-cluster primitive (heiman/eve/etc.) — that's a generator quirk independent of this change.

packages/custom-clusters/src/clusters/tcl.ts
iamadamreed: Verified by live read on H50D44W firmware 1.0:

```
1/322239491/5 = '[]'      (errorCodes — empty list as text)
1/322239491/6 = '[3]'     (featureSet)
```

The device serialises both attributes to text and emits them as Matter `string` (length-prefixed UTF-8 octet string), not as Matter list TLVs. Modeling them as `array[uint16]` would tell matter.js to expect a list TLV header that doesn't exist on the wire and would break reads. So `string` is the correct wire type; the JSON-list shape is application semantics that consumers parse on receipt.

Strengthened the JSDoc on both attributes in `d7a3632` to make the JSON-string-not-TLV-list distinction explicit. Full diagnostic is in the [comment above](https://github.com/matter-js/matterjs-server/pull/630#issuecomment-4416017065).

packages/custom-clusters/src/clusters/tcl.ts
iamadamreed: Yes — verified on the wire (live read returns the literal text `'[]'` / `'[3]'`, not a list TLV). `string` matches what the device emits. JSDoc tightened in `d7a3632`.

packages/custom-clusters/src/clusters/tcl.ts
iamadamreed: Same as the previous thread — wire format is a literal JSON-encoded string (`'[3]'` on H50D44W), not a list TLV. Static across observed captures. JSDoc cross-links to `errorCodes` via `{@link errorCodes}` in `d7a3632`.

packages/custom-clusters/src/clusters/tcl.ts
iamadamreed: Dropped in `d7a3632`. You're right — declaring a cluster whose contents we can't interpret adds noise without signal. The cluster still appears in the device's `Descriptor.ServerList`, but matter.js's default handling is enough for that; if/when someone identifies what the vendor-prefixed attribute contains, it can come back as a follow-up PR with the right type.

Apollon77:
@iamadamreed Can you please just attach the plain diagnostic JSON file ... I would like to see the raw values and not values post-processed by AI :-) Thanks a lot

iamadamreed:
@Apollon77 — sorry about that, fair point. Plain dump, straight from Home Assistant's matter integration diagnostics export (no post-processing, only the device serial redacted):

**https://gist.github.com/iamadamreed/e96a537639f5bbddf75085d612783995**

The TCL device is the only node in `data.server.nodes`; cluster `322239491` (`0x1334FC03`) attributes are at the bottom of its `attributes` map.

iamadamreed:
### Empirically verified error code semantics

Followed up on the earlier diagnostic with a controlled bucket-full / bucket-empty cycle on the same H50D44W. Goal: identify which attribute(s) actually carry the bucket-full signal.

Polled every attribute on the device every 3 s through the transition. Across a complete fill → empty → reinsert cycle, **only one attribute flips deterministically with bucket state**:

```
1/322239491/5  errorCodes        "[5]"  →  "[]"     ← THE bucket-full signal
1/322239491/2  currentHumidity     52   →   51      (~1% measurement noise, unrelated)
```

Things that did **not** change:

- `1/322239491/3` `waterBucketFull` — stayed `False` throughout. **This dedicated bool is dead on this device.**
- `1/322239491/4` `filterAlert` — stayed `False`
- `1/6/0x13340000` and `1/6/0x13340001` (TCL vendor extensions on the standard OnOff cluster) — stable, unrelated to bucket state
- FanControl `FanMode` / `PercentSetting` / `PercentCurrent` — stable

So: on the H50D44W, **error code 5 in `errorCodes` is the only reliable bucket-full signal**. The dedicated `waterBucketFull` bool at attr 3 looks like dead firmware on TCL's side.

Pushed `ce78ac9` on the PR branch to capture this in the JSDoc on `errorCodes`. Downstream consumer ([iamadamreed/ha-tcl-matter v0.4.2](https://github.com/iamadamreed/ha-tcl-matter/releases/tag/v0.4.2)) now reads code 5 in its bucket-full binary sensor, defensively OR'd with attr 3 so any future firmware that wires up the bool still works.

### Diagnostic artefacts

Added the empirical before/after dump to the existing gist alongside the original full diagnostic. All files are raw HA matter diagnostics output, only the device serial is redacted:

**https://gist.github.com/iamadamreed/e96a537639f5bbddf75085d612783995**

| File | What |
|---|---|
| `matter_diag_raw_redacted.json` | Original full HA matter diagnostics dump (from the earlier comment) |
| `tcl-h50d44w-bucket-cycle-diff.json` | Compact diff of just the attributes that changed across the bucket cycle |
| `tcl-h50d44w-bucket-cycle-01-tank-full.json` | Full node-attribute snapshot taken with the bucket physically full |
| `tcl-h50d44w-bucket-cycle-02-tank-emptied.json` | Same snapshot after the bucket was emptied + reinserted |

Diff is reproduced inline above; the two full snapshots are there if you want to verify the "everything else didn't change" claim against unprocessed values.

Apollon77:
Ok, yes all makes sense, thanks a lot

PR #633, opened 2026-05-10
Initial Push for Matter 1.5.1

Apollon77:
This PR updates matter,js to the latest nightly bringing Matter 1.5.1 support including strict attestation validation as it was in the Python server.
Also the Python classes got regenerated.

This is the basis for 0.7.x


python_client/matter_server/client/models/device_types.py
```diff
@@ -68,6 +68,19 @@ class AirQualitySensor(DeviceType):
     }
 
 
+fully ok class AudioDoorbell(DeviceType):
```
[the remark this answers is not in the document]
Apollon77: fixed

python_client/chip/clusters/cluster_defs/BasicInformation.py
```diff
@@ -72,11 +72,11 @@ def descriptor(cls) -> ClusterObjectDescriptor:
     serialNumber: typing.Optional[str] = None
     localConfigDisabled: typing.Optional[bool] = None
     reachable: typing.Optional[bool] = None
-    uniqueID: str = ""
+    uniqueID: typing.Optional[str] = None
```
TheJulianJES, line 75: Note for me/others: With `uniqueID` now being optional and no longer defaulting to an empty string, we may want to adjust typing for `unique_id` in the HA Matter integration here, or possibly default to an empty string in the integration: [homeassistant/components/matter/models.py#L37](https://github.com/home-assistant/core/blob/9b8c81cba1999fe437cf2fadb2629d332c7fc607/homeassistant/components/matter/models.py#L37)

At the moment, it's only used for the Google Home integration to populate some extra metadata. It doesn't handle `unique_id=None` properly, from what I can see, so we'll need to address that in HA. 

We may just want to return `None` in `get_matter_device_info` instead (if the `unique_id` is `None`): [homeassistant/components/matter/__init__.py#L51-L66](https://github.com/home-assistant/core/blob/9b8c81cba1999fe437cf2fadb2629d332c7fc607/homeassistant/components/matter/__init__.py#L51-L66)

In general, I don't think a device should ever have a `None` `uniqueID`?

python_client/chip/clusters/cluster_defs/Thermostat.py
```diff
@@ -32,7 +32,7 @@ def descriptor(cls) -> ClusterObjectDescriptor:
                 ClusterObjectFieldDescriptor(Label="absMaxCoolSetpointLimit", Tag=0x00000006, Type=typing.Optional[int]),
                 ClusterObjectFieldDescriptor(Label="PICoolingDemand", Tag=0x00000007, Type=typing.Optional[uint]),
                 ClusterObjectFieldDescriptor(Label="PIHeatingDemand", Tag=0x00000008, Type=typing.Optional[uint]),
-                ClusterObjectFieldDescriptor(Label="HVACSystemTypeConfiguration", Tag=0x00000009, Type=typing.Optional[Thermostat.Bitmaps.HVACSystemTypeBitmap]),
```
TheJulianJES, line 35: I've not looked at the Matter spec, but I'm guessing it's expected that the `HVACSystemTypeBitmap` is removed entirely (and this is typed as an `int` again)? It doesn't affect HA anyway.

TheJulianJES (approved):
From a quick glance, the Python cluster definition changes look good. Home Assistant Core still starts and loads the Matter integration as expected. Tests also pass.

There is only one thing with the optional typing change (`uniqueID: str = ""` changed to `uniqueID: str | None = None`), which may not be handled properly be the Google Home integration in Home Assistant if we actually get a `None` `unique_id`. Either way, this is something we can/should fix in HA, I think.

I've only looked at the Python stuff, but I've let Claude review the entire PR quickly. Its notes just copy-pasted (might be wrong, but probably still worth a look). Posted in a follow-up comment below.

TheJulianJES:
As noted in my review above, here's a "summary" of Claude reviewing this PR. The first two issues are related to the Copilot comments and follow-up fixes by Claude (Sonnet 4.6). This may be wrong, but it's likely still worth a look.

<details><summary>Claude review of entire PR (CLICK TO EXPAND/OPEN)</summary>
<p>

## Review notes

Tried this against Home Assistant core's Matter integration with the regenerated `matter-python-client` installed locally — all 365 HA Matter tests pass. A few things I'd like to flag before merging:

### 🟠 `DclBehavior` is no longer registered on the controller, but `agent.get(DclBehavior)` is still used

`packages/ws-controller/src/controller/MatterController.ts` removed:

```ts
this.#controllerInstance!.node.behaviors.require(DclBehavior, {
    fetchTestCertificates: this.#enableTestNetDcl,
});
```

…and replaced it with `new DclCertificateService(this.#env.root, …)` at line 186. But three accessors still do `agent.get(DclBehavior)…` (lines 264 / 277 / 290), and `vendorInfoService()` + `certificateService()` are pushed unconditionally onto `initPromises` whenever `commandHandler` starts (lines 232–233).

In matter.js, `Agent.get(Type)` ends up in `Behaviors.#createBacking` → `#getBehaviorType`, which throws `BehaviorInitializationError: Unsupported behavior` if the behavior wasn't explicitly `require()`'d (`@matter/node/src/endpoint/properties/Behaviors.ts:792-795`).

`DclBehavior` does get implicitly required by `SoftwareUpdateManager` (`@matter/node/src/behavior/system/software-update/SoftwareUpdateManager.ts:161`), but **only when `enableOtaProvider: true`** and the OTA provider endpoint is actually instantiated. With `--disable-ota-provider` nothing else registers it, so the first call to `vendorInfoService()` / `certificateService()` / `otaUpdateService()` will throw at runtime. Even on the default OTA-enabled path there's a timing concern about whether `SoftwareUpdateManager` has eagerly required `DclBehavior` by the time those init promises run.

This is what Copilot flagged on `MatterController.ts:185`, and it was not addressed in the `e6ad266` "fix PR review findings" commit. Cleanest fix is either:

- Restore `node.behaviors.require(DclBehavior, { fetchTestCertificates: this.#enableTestNetDcl })`, or
- Switch the three accessors to look up the services directly via `this.#env.root.get(DclCertificateService)` / `DclVendorInfoService` / `DclOtaUpdateService` (and instantiate the latter two in `initialize()` alongside the existing `DclCertificateService` at line 186).

### 🟡 `onAttestationFailure` logging is only half-fixed

`packages/ws-controller/src/controller/ControllerCommandHandler.ts:828-838`:

```ts
onAttestationFailure: findings => {
    let accept = true;
    for (const f of findings) {
        if (f.level === "error") {
            accept = false;
        }
        logger.info(`Attestation finding (${f.level}):`, f.type, f.message);
    }
    logger.info(`Attestation ${accept ? "accepted" : "rejected"}`);
    return accept;
},
```

The "single final accept/reject log after the loop" part of Copilot's feedback was applied, but `level` is `"error" | "warning" | "info"` (`@matter/protocol/src/certificate/DeviceAttestationValidator.ts:39`) and the loop still logs every finding at `info`, hiding severity. Should map to `logger.error` / `logger.warn` / `logger.info` per `f.level`.

Also worth a second look: the PR description says *"strict attestation validation as it was in the Python server"*, but this handler only rejects on `error` findings and silently accepts every `warning`. If "strict" was meant to mirror python-matter-server's stricter policy, this likely needs to fail on warnings too — worth confirming against the Python implementation.

### 🟢 Spec/typing changes that look intentional but are worth flagging

- **`Thermostat.OccupiedSetback{,Min,Max}` / `Unoccupied…`**: `Union[None, Nullable, uint]` → `Optional[uint]`. matter.js's model now marks these `conformance: "D"` (deprecated) with no type/quality, so the generator drops `Nullable`. Real Matter 1.3 devices in the wild still report `null` for these. In practice both `Nullable` and `None` round-trip to Python `None`, and HA already handles `None` correctly — but downstream consumers expecting the `Nullable` sentinel should be aware.
- **`BasicInformation.uniqueID`**: `str = ""` → `Optional[str] = None`. Matches the Matter spec (UniqueID is optional). On the HA side, this means `MatterDeviceInfo.unique_id` can now be `None`, which the Google Assistant integration forwards as `matterUniqueId: null` in the SYNC payload. We'll patch the HA side; just noting in case other downstream consumers are similarly affected.
- **`BasicInformation.specificationVersion` / `maxPathsPerInvoke`**: `uint = 0` → `Optional[uint] = None`. Same pattern — spec-correct, but consumers that read these without a `None` guard will need updating. (HA doesn't read them.)
- **`Globals.Enums.namespace`** has many tags renamed from `kX` to `kCommonX` (e.g. `kClosure` → `kCommonClosure`) and a new `kClosure` was assigned to `0x44`. Anyone referencing these by name (rather than tag ID) will break silently. HA doesn't, but worth noting in the changelog.
- **`ColorControl.colorCapabilities`**: type narrowed from `Globals.Bitmaps.map16` to `ColorControl.Bitmaps.ColorCapabilitiesBitmap`. HA already uses the new bitmap, so no impact there.

### 🟢 What was fixed correctly

- `fully ok class AudioDoorbell(DeviceType):` syntax error — confirmed clean.
- Duplicate `ColorCapabilitiesBitmap` class in `ColorControl.py` — collapsed to one (generator now skips model bitmaps that `EXTRA_BITMAPS` overrides).
- `teh` → `the` typo in `MatterController.ts`.

### TL;DR

Two concrete items I'd want addressed before merge:

1. **Fix the `DclBehavior` / `agent.get` mismatch** — at minimum verify the `--disable-ota-provider` path doesn't throw on first `commandHandler` start.
2. **Use proper log levels in `onAttestationFailure`** and confirm the accept-on-warning policy matches python-matter-server.

Everything else is non-blocking from my side.


</p>
</details>

Apollon77:
@TheJulianJES Ok the claude review is ok /partially wrong, so nothing to take from there.
DclBehavior: Works because the init of the behavior is sync and so "get" will work. The require basically did never work and never set the params provided because wa salready existing.
Logleves are intended that way.

python_client/chip/clusters/cluster_defs/BasicInformation.py
Apollon77: uniqueId was added in Matter 1.3 or such, so older devices never have one ... soooo it was "optional" anyway. event that the python layer maybe made that "empty strung fallback"?

python_client/chip/clusters/cluster_defs/Thermostat.py
Apollon77: HVACSystemTypeConfiguration was basically a "Zigbee derived value" and was e.g. "D" (deprecated) in Matter 1.4. 2 already ... so thats a complete removal now

Issue #659, opened 2026-05-18
Healthcheck fails when LISTEN_ADDRESS is set


jamesonuk:
### Version

v0.7.0

### Node.js Version

Per container

### Operating System

Docker

### Issue Description

Trivial issue but if you set `LISTEN_ADDRESS` environment variable then health check fails.

health check is configured to call localhost
https://github.com/matter-js/matterjs-server/blob/621bab3fb9465c2f223ebe79aad202c012738773/docker/matterjs-server/Dockerfile#L82
but with `LISTEN_ADDRESS` set the server is only running on that address not localhost.

### Steps to Reproduce

Setup in docker and set LISTEN_ADDRESS.
Container will show as unhealthy.

### Server Logs

```shell
Docker logs

    {
      "Start": "2026-05-18T15:38:42.717526092Z",
      "End": "2026-05-18T15:38:42.784692749Z",
      "ExitCode": 1,
      "Output": "  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current\n                                 Dload  Upload   Total   Spent    Left  Speed\n\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\r  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0\ncurl: (7) Failed to connect to localhost port 5580 after 0 ms: Could not connect to server\n"
    }
```

### Device Information

N/A

### Additional Context

Even less of an issue but perhaps adding `--no-progress-meter` to the curl command would make health check logs look better

### Pre-submission Checklist

- [x] I am running an Alpha/Beta version of the Matter.js server
- [x] I have included the complete server logs (not truncated)
- [x] I have checked existing issues to avoid duplicates

Apollon77:
I agree, but we need to ensure to use the first address because in fact listenaddress can be repeated and is basically a "list" :-)

markvp:
@Apollon77 I wonder if it's worth testing *all* listed addresses as a form of validation of the input...
"you've configured me to listen on address x.x.x.x but i'm not accessible on that address..."

Apollon77:
Thats not what the healthcheck is for and effectively software will bind and work on all or none, so in my yes not needede

markvp:
Fair enough, I thought it might just pick up accidental typos if they're in there.

Apollon77:
should be solved in 0.7.1 from yesterday

PR #661, opened 2026-05-19
fix(docker): healthcheck honours LISTEN_ADDRESS / PORT

Apollon77:
## Summary
- Replace hardcoded `curl http://localhost:5580/health` in the container healthcheck with a small `healthcheck.sh` that picks the first address from `LISTEN_ADDRESS` (falling back to `localhost`), uses `PORT`, brackets IPv6 literals, and passes `--no-progress-meter`.
- Same fix applied to `Dockerfile.dev` (which additionally was hitting `/` instead of `/health`).
- `docs/docker.md` documents the new behavior and the one limitation: interface names (e.g. `eth0`) cannot be resolved from the shell, so set `LISTEN_ADDRESS` to a literal IP — or use the repeatable `--listen-address` CLI flag and include a literal loopback alongside — to keep the healthcheck working.

Addresses #659.

### DIFF
--- a/docker/matterjs-server/Dockerfile
+++ b/docker/matterjs-server/Dockerfile
@@ -77,9 +77,13 @@ VOLUME ["/data"]
 # WebSocket API port
 EXPOSE 5580
 
+# Healthcheck script honours LISTEN_ADDRESS / PORT so it still works when the
+# server is bound to a non-localhost address. See script for limitations.
+COPY --chmod=0755 healthcheck.sh /usr/local/bin/healthcheck.sh
+
 # Health check - verify the server is responding
 HEALTHCHECK --interval=60s --timeout=10s --start-period=30s --start-interval=5s --retries=3 \
-    CMD curl -f http://localhost:5580/health || exit 1
+    CMD ["/usr/local/bin/healthcheck.sh"]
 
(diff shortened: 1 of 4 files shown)


Apollon77:
@jamesonuk makes that solution sense this way?

jamesonuk:
> @jamesonuk makes that solution sense this way?

It is a pretty trivial issue and you probably have more important issues to worry about 😄 

Makes sense and would seem to work.

I had a quick play and a couple of suggestions (but this is getting to the point of rolling it in glitter).
With the version of curl that ships in the container you don't need `--no-progress-bar` (I think this was a bit temperamental in some older versions.)   `-fsS` is sufficient to hide the progress bar.   You do get the valid response back so I would probably add `-o /dev/null` to disregard the actual response.

I would also add `-m 2` (or `--max-time 2`) or some other sensible timeout to stop it ever getting stuck.

Apollon77:
Thanks a lot!

schildbach:
I think this change is the wrong way. In Docker, if you want custom ports you use a port mapping.

jamesonuk:
> I think this change is the wrong way. In Docker, if you want custom ports you use a port mapping.

This is isn't ports, this is IP address / adapter

PR #662, opened 2026-05-19
feat(ble-proxy): end-to-end BLE proxy commissioning + matter-ble-proxy Python lib

Apollon77:
## Summary

End-to-end BLE proxy commissioning support, plus a brand-new
`matter-ble-proxy` Python client library that the Home Assistant Matter
integration consumes (see [companion HA core PR](https://github.com/home-assistant/core/pulls?q=is%3Apr+matter-ble-proxy)).

This is two logical changes squashed onto one branch:

1. **End-to-end BLE proxy commissioning** — Use an external BLE proxy to commission Matter devices directly into the Matter server. This lifts the limitation to only support local ble adapters. Matter device commissioning over the proxy now completes end-to-end (PASE → CASE → operational).

2. **`matter-ble-proxy` Python package** — new top-level `python_ble_proxy/` package. Ships protocol logic + Bleak-based default backend + CLI. Used by the HA Matter integration to integrate with this new BLE proxy feature of the matter server, and provides a standalone reference CLI for testing.

## matter server BLE Proxy support
- When enabled with `--ble-proxy`CLI option it supports a dedicated BLE proxy websocket protocol at the /ble path of the matter server. Details see https://github.com/matter-js/matterjs-server/blob/590e85dcfa560d03c55f3f3db2aeaf3efd74433c/docs/ble-proxy-protocol.md

## JavaScript example using noble
- CLI example including an example in JS which implements a ble-proxy client using "noble" to do the BLE logic and interacts with the server

## matter-ble-proxy Python library

- Public API: \`MatterBleProxy\`, \`BleScanSource\` /
  \`BleDeviceResolver\` ABCs, \`AdvertisementData\` dataclass,
  \`BleakScanSource\` / \`BleakDeviceResolver\` defaults.
- Resource-optimized: scanner only runs between
  matter-server's \`start_scan\` / \`stop_scan\`. Fully torn down on
  proxy disconnect. No idle scanning.
- All command handlers surface spec-defined error codes
  (\`read_failed\`, \`write_failed\`, \`subscribe_failed\`,
  \`not_subscribed\`, …) instead of bucketing into
  \`internal_error\`.
- INFO-level command logging mirroring the JS \`noble-ble-proxy\`
  example (\`[←CMD] id=… command …\`), with base64 payloads stripped
  for readability.
- \`matter-ble-proxy\` console entry: same flag set as the JS
  reference (\`--server\`, \`--log-level\`).
- Package is published with the same versioning in sync with the matter server

## Docs

- Top-level \`README.md\` — added "BLE Proxy mode" section pointing
  at both reference clients + the protocol spec.
- \`packages/ble-proxy/README.md\` — links to the Python package.
- \`python_ble_proxy/README.md\` — install, CLI, library API.
- \`docs/ble-proxy-protocol.md\` — new UUID-format section, clarified
  \`response\` field on \`write_characteristic\`, and the ATT Write
  Request requirement on \`WRITE_DATA\` binary frames.


python_ble_proxy/matter_ble_proxy/example/__init__.py
TheJulianJES, line 1: I'm not sure the `example` should be part of `matter_ble_proxy`. But if we always want to include some basic CLI, then `cli` may be a better name than `example`.

PR #715, opened 2026-06-03
README.md: add Home Assistant and ESPHome requirements to BLE proxy description

schildbach:


README.md
```diff
@@ -127,6 +127,22 @@ Home Assistant's Matter integration, the Bleak-based Python CLI
 bridges BLE traffic from wherever the adapter actually lives. See
 [`docs/ble-proxy-protocol.md`](docs/ble-proxy-protocol.md) for the protocol spec.
 
+For BLE proxy mode to work, you need:
```
Apollon77:
```suggestion
When using the BLE proxy mode together with Home Assistant, you need:
```

README.md
```diff
@@ -127,6 +127,22 @@ Home Assistant's Matter integration, the Bleak-based Python CLI
...
+
+- Home Assistant 2026.6 (or later)
+- ESPHome 2026.5 (or later)
+- esp-idf 6.0.1 (exactly)
```
Apollon77: why exactly and not "or higher"?

schildbach: Because at the moment the necessary fix is only on 6.0.1. We can adapt the wording once the fix gets ported to other branches. (I hope it will be ported to 5.5.x, the current default.)

README.md
schildbach: Is there a way to use BLE proxy *without* Home Assistant? It's my understanding that this is not possible, as BLE proxies are a feature of HA.

Apollon77: The Matter server implements a generic BLE Websocket interface and a python client to connect "python bluetooth logic to that". HA then has logic to integrate the ESPHome BLE Proxy feature with that python bluetooth stuff. SO the server is generic here

README.md
Apollon77: ok but it will also stay in 6.02 I guess :-) so I think 6.0.1+ is the better wording?

schildbach: Well, there is no 6.0.2 yet. Only a 6.1-dev, and the fix isn't in there. On top of that, ESPHome at the moment only supports 6.0.1 not 6.0.2 or 6.1-dev. So I think until the situation is cleared up, 6.0.1 is the better wording. We can change it later.

README.md
schildbach: Ok, I applied the change.

schildbach:
Rebased on current main. If you agree with the 6.0.1 wording, this is ready to merge.

README.md
```diff
@@ -127,6 +127,22 @@ Home Assistant's Matter integration, the Bleak-based Python CLI
 bridges BLE traffic from wherever the adapter actually lives. See
 [`docs/ble-proxy-protocol.md`](docs/ble-proxy-protocol.md) for the protocol spec.
 
+When using the BLE proxy mode together with Home Assistant, you need:
+
+- Home Assistant 2026.6 (or later)
+- ESPHome 2026.5 (or later)
+- esp-idf 6.0.1 (exactly)
```
Apollon77:
```suggestion
- esp-idf 6.0.1 (or later)
```

README.md
```diff
@@ -127,6 +127,22 @@ Home Assistant's Matter integration, the Bleak-based Python CLI
 bridges BLE traffic from wherever the adapter actually lives. See
 [`docs/ble-proxy-protocol.md`](docs/ble-proxy-protocol.md) for the protocol spec.
 
+When using the BLE proxy mode together with Home Assistant, you need:
```
Apollon77:
```suggestion
When using the BLE proxy mode together with Home Assistant (which uses the ble-prox-python -client), you need:
```

README.md
```diff
@@ -127,6 +127,22 @@ Home Assistant's Matter integration, the Bleak-based Python CLI
 bridges BLE traffic from wherever the adapter actually lives. See
 [`docs/ble-proxy-protocol.md`](docs/ble-proxy-protocol.md) for the protocol spec.
 
+When using the BLE proxy mode together with Home Assistant (which uses the ble-prox-python -client), you need:
```
Apollon77:
```suggestion
When using the BLE proxy mode together with Home Assistant (which uses the ble-proxy-python -client), you need:
```

Trinaj037 (approved):

schildbach:
esp-idf 6.0.2 (recently released) has the fix as well, but the current version of ESPHome does not allow to use that idf version.

I still think we should omit the wording "or later" from the README, until esp-idf 6.x becomes the default for ESPHome.

Trinaj037 (approved):

PR #756, opened 2026-06-17
feat: Matter multicast group support (controller + WS API)

cedricziel:
Standards-compliant Matter **multicast group** support: one command drives all members via real IPv6 groupcast (not unicast fan-out). Server-side half of "let HA consume groups, map them as entities, command a group's members". OHF extension — superset of the Python Matter Server API (no group commands; python-matter-server#1071). Groupcast can't ride the unicast-only `device_command`, and the operational group key is controller-fabric state, so these are first-class commands. Feature-detect via `ServerInfoMessage.groups_supported`.

**Commands** (`ws-client`): `create_group`, `delete_group`, `get_groups`, `add_group_member`, `remove_group_member`, `send_group_command`, `reconcile_group`; `group_added/updated/removed` events.

**`GroupManager` + `GroupRegistry`** (`ws-controller`): installs the operational group key on the controller's own fabric (matter.js doesn't persist `FabricGroups` across restarts — re-installed on boot from the registry; epoch key never leaves the server), provisions members over unicast (`KeySetWrite` → `GroupKeyMap` → `Groups.AddGroup` → group ACL), and sends groupcasts via the matter.js `ClientGroup` path.

**Verified against Matter 1.4 Core** (CSA 23-27349): group-id range 0x0001–0xFEFF (§2.5.4), keyset 0 = IPK (§4.17.3.5.1), ACL Group subject = uint64 low-16 = GroupId + never Administer (§9.10.5.6/§6.6.6.2), EpochStartTime epoch-µs, only 0 illegal (§11.2.7.1), TrustFirst mandatory (§11.2.5.1).

**Hardening:** mutex-serialized mutations; ACL/`GroupKeyMap` RMW from a fresh fabric-filtered read (never drops the admin ACL entry → no controller lock-out); surfaces failed `AddGroupResponse` status; optional least-privilege ACL `targets`; `reconcile_group` checks registry vs each device's `GroupKeyManagement.GroupTable` and repairs drift.

**Tests:** format/build/lint ✓, 121/121 unit ✓ (26 new). ⚠️ groupcast delivery **not yet hardware-verified** — needs a real LAN with IPv6 multicast (matter.js has no mock-multicast loopback).

**Follow-ups (separate):** HA-core entity integration; dashboard group UI.


cedricziel:
@Apollon77 - I intentionally marked this as draft as I wanted to get some feedback on the design here. I think there's some headroom to make groups manageable for consumers. - yes a client like HA can jump through all hoops and make  create the key, provision it onto devices etc etc - but this PR would make it a ton easier and open opportunities for another ui in the dashboard. looking forward to your feedback :)

Apollon77:
@cedricziel Thanks a lot for this contribution and in fact yes this is a big part of what I also have in mind (just in a different place ;-) ) and started working on it. but because this is bigger it needs a bit of time and we could start with this but then in 1-2 months remove big parts again when the real system included in matter.js is done. ... I also have locally some "Matter dashboard UI experiments" and how to expose it to where and how to configure it for end users is even a "different "thing to discuss then with the Home Assistant guys (including what's best where).

Can you contact me in discord for a sync? Might be easier than here.

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+const GROUP_TABLE_ATTRIBUTE_ID = GroupKeyManagement.attributes.groupTable.id;
+
+/**
+ * Operations the {@link GroupManager} performs against individual member nodes.
+ *
+ * Implemented by {@link ControllerCommandHandler}, which owns the node interaction
+ * primitives. Values passed/returned here are native (already Matter-shaped), not
+ * the WebSocket tag-based representation.
+ */
+export interface GroupMemberOps {
```
Apollon77, line 48: Honestly, no need to map that to the matterjs-server logic that always adds an additional layer. the Group manager has the commissioningcontroller and the nodes available and can do the writes directly use await node.setStateOf(ClusterType, { name: value})

packages/ws-client/src/models/model.ts
```diff
@@ -295,6 +295,63 @@ export interface APICommands {
...
+    create_group: {
+        requestArgs: {
+            name: string;
+            /** Optional explicit application group id (0x0001..0xFEFF). Auto-allocated when omitted. */
+            group_id?: number;
+            /**
+             * Optional ACL targets scoping the group's access on each member (least
+             * privilege). When omitted the group is granted Operate on all clusters.
+             */
+            acl_targets?: AccessControlTarget[];
```
Apollon77, line 318: acl targets already include the endpoint  and cluster? they not really belong into the group record unless you want to pre-prime a group for a dedicated use case which is problematic anyway because of the endpoint Id ... and would only work for clusters.

I see that the code uses this to add acls for the joined member but I do not really get the real idea of that.

packages/ws-client/src/models/model.ts
```diff
@@ -295,6 +295,63 @@ export interface APICommands {
...
+    };
+    delete_group: {
+        requestArgs: { group_id: number };
+        response: null;
+    };
+    get_groups: {
+        requestArgs: Record<string, never>;
+        response: MatterGroupData[];
+    };
+    add_group_member: {
```
Apollon77, line 330: lets change to "join_group" because this way it is also called in matter and will later on be with new Groupcast coming

packages/ws-client/src/models/model.ts
```diff
@@ -295,6 +295,63 @@ export interface APICommands {
...
+        requestArgs: { group_id: number; node_id: number | bigint; endpoint_id: number };
+        response: MatterGroupData;
+    };
+    remove_group_member: {
```
Apollon77, line 334: And here "leave_group"

packages/ws-client/src/models/model.ts
```diff
@@ -295,6 +295,63 @@ export interface APICommands {
...
+        requestArgs: { group_id: number; node_id: number | bigint; endpoint_id: number };
+        response: MatterGroupData;
+    };
+    send_group_command: {
```
Apollon77, line 338: formally not needed because each group has it's own Matter node Id which contains the groupId, so I would favor to not have a special websocket command to send a command but just use the normal one with a group node id provided

packages/ws-client/src/models/model.ts
```diff
@@ -335,6 +392,55 @@ export interface AccessControlTarget {
...
+ *
+ * The operational epoch key itself is sensitive and is never exposed over the API.
+ */
+export interface MatterGroupData {
+    /** Matter group id (1..0xFFFF). */
+    group_id: number;
+    /** Human-readable group name (<=16 chars, per the Groups cluster). */
+    name: string;
+    /** Group key set id binding the group to its operational key on the fabric. */
+    group_key_set_id: number;
```
Apollon77, line 416: why this is needed for externals?

packages/ws-client/src/models/model.ts
```diff
@@ -335,6 +392,55 @@ export interface AccessControlTarget {
...
+    /** Devices that have been provisioned into the group. */
+    members: MatterGroupMember[];
+    /**
+     * ACL targets scoping the group's access on each member. When absent the group
+     * is granted Operate on all clusters/endpoints.
+     */
+    acl_targets?: AccessControlTarget[];
```
Apollon77, line 423: see above

packages/ws-client/src/models/model.ts
```diff
@@ -384,6 +490,8 @@ export interface ServerInfoMessage {
     ble_proxy_enabled?: boolean;
     /** The controller's own operational (CASE) node id. Note: Only available with OHF Matter Server, not Python Matter Server. */
     controller_node_id?: number | bigint;
+    /** True when the server supports Matter multicast group commands (`create_group`, `send_group_command`, …). OHF Matter Server extension; absent on Python Matter Server. */
+    groups_supported?: boolean;
```
Apollon77, line 494: This is a feature additional that usually means we increase the websocket schema version and via this users know which kind of protocol it is and what is supported. So this would be the right way to do that. And because this is just "adding commands" the server can easiely support the current and this next version because they are not really breaking.

packages/ws-client/src/models/model.ts
```diff
@@ -418,6 +526,15 @@ export interface APIEvents {
     webrtc_callback: {
         data: WebRtcCallbackData;
     };
+    group_added: {
```
Apollon77, line 529: whats the use case for the events?

packages/ws-controller/src/controller/GroupRegistry.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Backed by the same {@link StorageService} the rest of the controller uses, under
+ * a dedicated `groups` storage. Holds an in-memory map mirrored to disk so the
+ * controller can re-provision its own fabric key material after a restart.
+ */
+export class GroupRegistry {
+    #store: StorageContext;
+    readonly #groups = new Map<number, GroupRecord>();
+
+    static async create(env: Environment): Promise<GroupRegistry> {
+        const storage = await env.get(StorageService).open("groups");
```
Apollon77, line 43: storage also need to be closed on shutdown - this is missing which means that the storage needs to given into the instance and the context should be created there and GroupRegistry should get a close method and wired into the close logiv

packages/ws-controller/src/controller/GroupRegistry.ts
```diff
@@ -0,0 +1,145 @@
...
+        await instance.#load();
+        return instance;
+    }
+
+    private constructor(store: StorageContext) {
+        this.#store = store;
+    }
+
+    async #load() {
+        const records = (await this.#store.get("groups", [])) as unknown as GroupRecord[];
```
Apollon77, line 55:
```suggestion
        const records = (await this.#store.get<GroupRecord[]>("groups", []));
```

... better typing

packages/ws-controller/src/controller/GroupRegistry.ts
```diff
@@ -0,0 +1,145 @@
...
+     * Lowest unused application Group ID (0x0001..0xFEFF), or throw if exhausted.
+     * The 0xFF00..0xFFFF range is reserved for universal groups and never allocated.
+     */
+    allocateGroupId(): number {
+        for (let candidate = 1; candidate <= 0xfeff; candidate++) {
+            if (!this.#groups.has(candidate)) {
+                return candidate;
+            }
+        }
+        throw new Error("No free application group id available (0x0001..0xFEFF exhausted)");
```
Apollon77, line 90: Please define dedicated errors transporting a meaning instead of plain Errors

packages/ws-controller/src/controller/GroupRegistry.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    /** Lowest unused group key set id (1.. ; 0 is reserved for the IPK, Matter 1.4 Core §4.17.3.5.1). */
+    allocateGroupKeySetId(): number {
+        const used = new Set([...this.#groups.values()].map(g => g.group_key_set_id));
+        for (let candidate = 1; candidate <= 0xffff; candidate++) {
+            if (!used.has(candidate)) {
+                return candidate;
+            }
+        }
+        throw new Error("No free group key set id available");
```
Apollon77, line 101: same here

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+    ServerError,
+} from "../types/WebSocketMessageTypes.js";
+import { GroupRecord, GroupRegistry, toPublic } from "./GroupRegistry.js";
+
+const logger = new Logger("GroupManager");
+
+/**
+ * Epoch start time for the single operational epoch key we provision. Units are
+ * epoch-us (microseconds since 2000-01-01 UTC, Matter 1.4 Core §11.2.5.4); a value
+ * of 1 keeps the key always-active. Per §11.2.7.1 only 0 is illegal in KeySetWrite.
```
Apollon77, line 27: "always active" is not really true but lets phrase it that way, using 1 for the first key of a keyset is ok

packages/ws-controller/src/controller/ControllerCommandHandler.ts
```diff
@@ -244,6 +252,26 @@ export class ControllerCommandHandler {
         logger.notice(`Matter Controller started`);
         this.#peers = this.#controller.node.env.get(PeerSet);
 
+        // Initialize multicast group support: load persisted groups and re-install
+        // their operational key material onto the controller's own fabric.
+        const groupRegistry = await GroupRegistry.create(this.#controller.node.env);
```
Apollon77, line 257: Why the groupregistry needs to be an entity out side here? Why the Groupmanager does not create the needed  registry itself and uses internally? And with GroupManager.create() you also preve the special "restrore" call here externally

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+     * otherwise collide — mirrors {@link ConfigStorage}'s node-id allocation mutex.
+     */
+    readonly #mutex = new Mutex(this);
+    readonly events = {
+        groupAdded: new Observable<[MatterGroupData]>(),
+        groupUpdated: new Observable<[MatterGroupData]>(),
+        groupRemoved: new Observable<[number]>(),
+    };
+
+    constructor(controller: CommissioningController, registry: GroupRegistry, ops: GroupMemberOps) {
```
Apollon77, line 112: see above. I would also do a create for GroupManger and incorporate the GroupRegistry here

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+            }
+            if (this.#registry.has(id)) {
+                throw ServerError.invalidArguments(`Group ${id} already exists`);
+            }
+
+            const epochKey = this.#controller.node.env.get(Crypto).randomBytes(GROUP_KEY_LENGTH);
+            const record: GroupRecord = {
+                group_id: id,
+                name,
+                group_key_set_id: this.#registry.allocateGroupKeySetId(),
```
Apollon77, line 152: see above this likely will not work that good when you consider devices being part of multiple groups

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+                members: [],
+                epoch_key_hex: Bytes.toHex(epochKey),
```
Apollon77, line 154: Why doing this conversion and later back to bytes? Just leave it Bytes type?

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+            // group can always be removed from the controller.
+            for (const member of [...record.members]) {
+                try {
+                    await this.#deprovisionMember(
+                        record,
+                        NodeId(BigInt(member.node_id)),
+                        EndpointNumber(member.endpoint_id),
+                    );
+                } catch (error) {
+                    logger.warn(`Failed to de-provision node ${member.node_id} from group ${groupId}:`, error);
```
Apollon77, line 188: is such a warn in the log enough to let a user know that he now has stale entries?

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+        });
+    }
+
+    /**
+     * Send a command to all members of a group via IPv6 group multicast.
+     *
+     * Groupcast is fire-and-forget: responses are suppressed at the protocol layer,
+     * so this resolves once the multicast has been emitted, not once members act.
+     */
+    async sendCommand(groupId: number, clusterId: number, commandName: string, payload?: unknown): Promise<void> {
```
Apollon77, line 285: see above, I would remove this here and add a short check in the normal invoke and write methods to detect group node ids and check group existence and such

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+    /**
+     * Run the full on-device provisioning sequence for one (node, endpoint): install the
+     * key set, map the group key, add the endpoint to the group, and grant the group ACL.
+     * Shared by {@link addMember} and {@link reconcileGroup}'s repair path.
+     */
+    async #provisionMember(record: GroupRecord, nodeId: NodeId, endpointId: EndpointNumber): Promise<void> {
+        const groupId = record.group_id;
+        const fabricIndex = this.#ops.currentFabricIndex(nodeId);
+
+        // 1. Install the operational key set on the member device.
```
Apollon77, line 384: node availability check missing

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+        await this.#ops.invokeNative(nodeId, EndpointNumber(0), GroupKeyManagement.id, "keySetWrite", {
+            groupKeySet: this.#buildGroupKeySet(record),
+        });
+
+        // 2. Map the GroupId -> KeySetId on the member (read-modify-write, fabric-scoped list).
+        const keyMap = (await this.#readList(nodeId, GroupKeyManagement.id, GROUP_KEY_MAP_ATTRIBUTE_ID)).filter(
+            entry => Number((entry as { groupId?: number }).groupId) !== groupId,
+        );
+        keyMap.push({ groupId: GroupId(groupId), groupKeySetId: record.group_key_set_id, fabricIndex });
+        await this.#ops.writeNative(nodeId, EndpointNumber(0), GroupKeyManagement.id, "groupKeyMap", keyMap);
```
Apollon77, line 394: any error response is ignored

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+
+        // 3. Add the endpoint to the group via the Groups cluster, and surface a failed
+        //    AddGroupResponse status (e.g. RESOURCE_EXHAUSTED) instead of silently succeeding.
+        const addResponse = await this.#ops.invokeNative(nodeId, endpointId, Groups.id, "addGroup", {
+            groupId: GroupId(groupId),
+            groupName: record.name,
+        });
+        const addStatus = Number((addResponse as { status?: number } | undefined)?.status ?? Status.Success);
```
Apollon77, line 402: if you call command directly using node.commandsOf(...).addGroup then also the typing fun goes away

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+        if (addStatus !== Status.Success) {
+            throw ServerError.sdkStackError(`AddGroup on node ${nodeId}/${endpointId} failed with status ${addStatus}`);
```
Apollon77, line 404: No already installed values are rolled back from above

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+
+        const next = acl.filter(entry => !this.#isGroupAclEntry(entry, groupId));
+        next.push({
+            privilege: AccessControl.AccessControlEntryPrivilege.Operate,
+            authMode: AccessControl.AccessControlEntryAuthMode.Group,
+            subjects: [GroupId(groupId)],
+            targets: desiredTargets,
+            fabricIndex,
+        });
+        await this.#ops.writeNative(nodeId, EndpointNumber(0), AccessControl.id, "acl", next);
```
Apollon77, line 491: write result ignored

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+            groupId: GroupId(groupId),
+            groupName: record.name,
+        });
+        const addStatus = Number((addResponse as { status?: number } | undefined)?.status ?? Status.Success);
+        if (addStatus !== Status.Success) {
+            throw ServerError.sdkStackError(`AddGroup on node ${nodeId}/${endpointId} failed with status ${addStatus}`);
+        }
+
+        // 4. Grant the group its ACL entry (read-modify-write).
+        await this.#ensureGroupAcl(record, nodeId, fabricIndex);
```
Apollon77, line 408: if anything fails here nothing rolls back the installed group

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+            m => BigInt(m.node_id) === BigInt(nodeId) && m.endpoint_id !== endpointId,
+        );
+        if (nodeHasOtherEndpoint) {
+            return;
+        }
+
+        const keyMap = (await this.#readList(nodeId, GroupKeyManagement.id, GROUP_KEY_MAP_ATTRIBUTE_ID)).filter(
+            entry => Number((entry as { groupId?: number }).groupId) !== groupId,
+        );
+        await this.#ops.writeNative(nodeId, EndpointNumber(0), GroupKeyManagement.id, "groupKeyMap", keyMap);
```
Apollon77, line 520: response ignored

packages/ws-controller/src/controller/GroupManager.ts
```diff
@@ -0,0 +1,527 @@
...
+
+        const acl = (await this.#readList(nodeId, AccessControl.id, ACL_ATTRIBUTE_ID)).filter(
+            entry => !this.#isGroupAclEntry(entry, groupId),
+        );
+        await this.#ops.writeNative(nodeId, EndpointNumber(0), AccessControl.id, "acl", acl);
```
Apollon77, line 525: too

packages/ws-controller/src/server/WebSocketControllerHandler.ts
```diff
@@ -632,6 +664,7 @@ export class WebSocketControllerHandler implements WebServerHandler {
             bluetooth_enabled: this.#commandHandler.bleEnabled,
             ble_proxy_enabled: this.#commandHandler.bleProxyEnabled,
             controller_node_id: this.#commandHandler.getCommissionerNodeId(),
+            groups_supported: true,
```
Apollon77, line 667: see above, better lets increase schema version but still support the old one ... and basically the python client needs also to be adjusted with the new version and code

Apollon77 (changes requested):
Ok, I did a first review pass over the PR and beside concrete comments on places here some general review feedback:

* The whole logic misses error handling, especially when multiple writes or commands are executed for one action it leaks entries that are not removed in case of the failure of an other one. This is especially problematic because of the next topic ...
* Node limits for ACLs (AccessControlEntriesPerFabric, TargetsPerAccessControlEntry, SubjectsPerAccessControlEntry) or GKM (MaxGroupsPerFabric, MaxGroupKeysPerFabric - we usually have a max of 4 groups per device per fabric and max 3 -no 2! -  group keys per device per fabric) are not checked or verified before trying to add new values. Group Limits are in 2.11.1.2. Group Limits and till Matter 1.6.0 most devices use these minimas as defaults. ACL limits are in 2.11.1.1. Additionally for sleepy ICD device the practical number of groups that will work with thread is 1 or maybe max 2. Access Control Limits and mean that (excluding the one Admin ACL) we have usually maximum 3 more ACL entries we can create
* With usual limits here also the idea to create one own group key set per group - which would be very good security wise - will not work with the matter device reality. So this requires a bit more thinking how to structure this or we need to start with exactly one group key set for all ... I need to think whats best

Yes I know also the current ACL logic does not check the size beforehand (I will add soon) and this was already the reason for failures with incomplete writes and so also different write success or error s with the matter.js server because here we write differently than python that also returned the write failures but noone ever checked that.

Additionally we need to adjust the python client code for the new command s and such (But can be done later when we are fine with the general structure and such).

We should also enhance the integration tests to really use them against the test device to have a practical proof

packages/ws-client/src/models/model.ts
extended-static: I agree with this comment. Using `send_device_command` instead of adding a `send_group_command` would be consistent with chip-tool's interface. With chip tool, unicast and multicast commands are more or less the same; group commands simply use the nodeid of the group and either omit or ignore the endpoint argument.

Usability is worth considering when UI elements are implemented. It's relatively straightforward to convert a hexadecimal group ID into a node ID, but somewhat more painful to do with decimal IDs.

extended-static:
I would be happy to test this on my deployment of HA once the initial peer review cycle is complete.

I've been working on group bindings with Inovelli as well.

I have 50 devices on my matter network. Group commands would help to reduce the amount of traffic on my network, and improve the user experience.

Apollon77:
@extended-static 
>  Group commands would help to reduce the amount of traffic on my network, and improve the user experience.

... they could (!) do this ... details depends highly on use cases, details of your network and other things

extended-static:
> ... they could (!) do this ... details depends highly on use cases, details of your network and other things

I'm planning to hack together a small add-on for HA that would allow Adaptive Lighting to send group commands instead of adjusting lights individually. The hack would also report the lights as being always on, using the `ExecuteIfOff` attribute to continuously adjust the bulbs, rather than performing an initial adjustment when they turn on.

Using groups could cut the number of messages AdaptiveLighting sends by about a third (in my home) and would reduce/mitigate the storm of messages that are generated when the lights initially turn on.

From my understanding, group bindings could allow the initial OnOff command to reach all of my devices before the flurry of response traffic. My lights felt very responsive when I initially tested group bindings using chip tool.

My user experience with unicast bindings is good, but not perfect. It can sometimes take a moment for all of my lights to turn on, and there is sometimes a bit of the popcorn effect. Seamless light control can be a sticking point when I try to promote home automation.

I should add: Further down the line, I'm hoping to see groups exposed to HA in a generic way (as devices or entities.)

Apollon77:
@extended-static Even if completely off topic here, but Adaptive lighting usually does not need many messages if you do it like e.g. apple with just "CT transitions with very long transition times" which then get updated over time ... with this you can end up with one message per light per hour or even less. Rething approaches please to generate the least possible messages. But better to discuss that in an own HA issue or Discord

extended-static:
> But better to discuss that in an own HA issue or Discord

I've filed a couple of feature requests that would address my specific problems.

https://github.com/orgs/home-assistant/discussions/4273
https://github.com/basnijholt/adaptive-lighting/issues/1495

Regardless, group bindings should help improve the responsively of my lighting. :)

PR #803, opened 2026-07-04
feat: Thread Network diagnostics, credential lists, and @matter/thread-br-client integration

Apollon77:
## Summary

Adds the **Thread Network diagnostics** feature and consumes the Thread Border Router stack from the published `@matter/thread-br-client` package (moved into the matter.js monorepo).

- **Thread Network diagnostics** — `get_thread_diagnostics` WS command + `thread_diagnostics_updated` event: per-network diagnostic batches from Border Routers over MeshCoP (CoAP/DTLS) or OTBR REST (auto-selected, cached). Rich read-only data (topology, link/MLE counters, battery, EUI64, stack version, Network Data, child IPv6, router-neighbor metrics) + counter reset + energy-scan / PAN-ID-conflict queries.
- **Dashboard Thread graph** renders the real mesh (route64 links + child counts) per network.
- **Named credential lists** — WiFi/Thread credentials as named entries (`id`); `get_all_credentials`; optional `id` on set/remove + `commission_with_code`; WS schema 12 (min 11, backward-compatible).
- **`--disable-thread-diagnostics`** flag (env `DISABLE_THREAD_DIAGNOSTICS`) disables the whole Thread BR subsystem for plain Matter-controller use; commissioning unaffected.
- Consumes `@matter/thread-br-client` (in-tree `@matter-server/thread-br` prototype removed).

## Verification

Full gate green: format / lint / build / full suite (incl. matter-server integration). Live-tested against real Border Routers (MeshCoP path across multiple networks). A cumulative independent adversarial review of the net diff passed (one cache-staleness bug found + fixed).

## ⚠️ Merge note — matter.js version reconciliation

This branch is on `@matter/*` **0.17.5-alpha** (which `@matter/thread-br-client` was built + tested against); `main` just released **0.17.4 stable** (1.1.7). Merging `main` conflicts only on the `@matter/*` version pins. Decision needed at merge time: align to 0.17.4 stable (verify thread-br-client compat) vs stay on 0.17.5-alpha. Not resolved here by design.

## Follow-up (upstream)

`@matter/thread-br-client`'s `OtbrRestProbe` rejects real HA OTBRs (`/node` lacks `numOfRouter`), so REST never engages and everything falls back to MeshCoP (works, but REST is the faster path). Tracked for the package.


Apollon77:
Merged latest `main` (kept `@matter/*` on the 0.17.5-alpha line — `@matter/thread-br-client` is built + tested against it) and addressed the review comments (4874878):

- **`get_thread_diagnostics` schema gating** (border-router-store `refresh` + `refreshDiagnosticsFor`): now requested with `require_schema=12` and error-isolated, so an older server or a transient diagnostics error no longer aborts the border-router list refresh (best-effort feature).
- **CLAUDE.md package list**: corrected — six packages, dropped the removed `thread-br` entry, added `custom-clusters`/`ws-client`/`ble-proxy`, and fixed the test-scoping examples.

Full gate green (format/lint/build + full suite incl. matter-server integration 248/248). Note: `@matter/*` intentionally stays on 0.17.5-alpha here.

Apollon77:
Addressed the latest review pass (5b241f0):

- **`get_thread_diagnostics` extPanId validation** ✅ — a malformed `extPanId` reached `Bytes.fromHex` and surfaced as a generic `UnknownError`. Now rejected with `ServerError.invalidArguments` (must be 16 hex chars) for a consistent `error_code`.
- **settings-dialog "stray `>`" (lines 347 / 453)** — assessed as a **false positive**. The template is `<md-text-button …>Clear</md-text-button\n>` — the trailing `>` on its own line is the closing bracket of `</md-text-button>` split across lines (the repo's lit whitespace style, same as the sibling `md-icon-button` blocks). Tags match and nothing renders a literal `>`, so no change.

Full gate green (build/lint + full suite incl. matter-server integration 248/248).

Apollon77:
Latest review pass (b5817c9):

- **Commission applies empty credentials** ✅ — now only applies a stored WiFi/Thread credential when its values are actually present (non-empty ssid+password / dataset); an empty stored entry is no longer pushed to the device.
- **ws-client over-gates the default id** ✅ — schema 12 is now required only for a non-default id (`id !== "default"`), matching the server + python clients. Passing the reserved `"default"` id no longer throws `InvalidServerVersion` against schema-11 servers.
- **force refresh re-probing REST (ThreadDiagnosticsService)** — **not changed; flagging as a design decision.** The current behavior is deliberate and covered by the test *"force=true does not re-probe a known REST-less BR (trusts the first-seen miss)"*: a force refresh re-pulls diagnostics but trusts the first-seen REST probe result; re-probing happens on mDNS re-announce, to avoid a ~1.5s REST-probe timeout on every manual refresh. The trade-off the reviewer raises is real (a *transient* probe miss stays until re-announce). Happy to flip it to "force re-probes REST" (and update that test) if you'd prefer recovery-on-force over the latency saving — your call.

Full gate green (build/lint + full suite incl. matter-server integration 248/248).

Apollon77:
Latest pass (14254f8):

- **`stop()` `certificateService()`** ✅ removed — good catch (and it matches `main`, which had already dropped it). It threw *"Controller not initialized"* when the controller never started (startup failed before init), breaking shutdown/cleanup.
- **force refresh now re-probes REST** ✅ — per maintainer decision, a `force=true` refresh re-probes (passing `force` through the `#probeAttempted` guard) so a transient earlier miss recovers instead of pinning the network to MeshCoP until the next mDNS re-announce. Updated the corresponding test.
- **no-arg `get_thread_diagnostics` returns `[]` vs `undefined`** — not changing: the no-arg form is a *list* endpoint (all networks), so an empty array is the correct "no networks" result and consumers handle it (`border-router-store` uses `Array.isArray(result)`); `undefined` is specifically the single-`extPanId` "nothing cached" signal (a different response shape by design).

Full gate green (build/lint + full suite incl. matter-server integration 248/248), on top of the latest `main` (Matter 1.6).

python_client/matter_server/common/models.py
```diff
@@ -24,6 +24,7 @@ class EventType(Enum):
     ENDPOINT_ADDED = "endpoint_added"
     ENDPOINT_REMOVED = "endpoint_removed"
     WEBRTC_CALLBACK = "webrtc_callback"
+    THREAD_DIAGNOSTICS_UPDATED = "thread_diagnostics_updated"  # schema 12+
```
TheJulianJES, line 27: Older library versions won't have this `EventType` type and I think they would currently crash if the Matter.js-Server still emits that event (for older/current `matter-python-client` versions + all `python-matter-server` versions used in way older HA Core versions). Should be checked.

If true, I guess we'd need to gate emitting the event per connection on the connected schema version.

I'm not sure this behavior is expected in the library though? Maybe something we should fix in another PR as well. Still, that'll only help for the latest library version then, used by whatever next HA Core release.

Apollon77:
@zigpy-review-bot Please re-review

PR #813, opened 2026-07-07
fix: convert ProvideOffer/SolicitOffer payload keys to matter.js naming

### LINKED ISSUE #812: ProvideOffer/SolicitOffer: webRtcSessionID payload key not mapped to matter.js's webRtcSessionId, causing ValidationMandatoryFieldMissingError
## Description

When a client (e.g. Home Assistant's Matter camera integration) sends a `ProvideOffer` command to `WebRtcTransportProvider` using the wire-format field names generated by the Python CHIP SDK client bundled in this repo (`python_client/chip/clusters/cluster_defs/WebRtcTransportProvider.py`), the command fails with:

```
ValidationMandatoryFieldMissingError/128: Missing mandatory field webRtcSessionId
```

even though the mandatory (nullable) field is present in the payload.

## Root cause

`ControllerCommandHandler.sendWebRtcProviderCommand` (`packages/ws-controller/src/controller/ControllerCommandHandler.ts`) builds the matter.js command `fields` by spreading the incoming `payload` object as-is:

https://github.com/matter-js/matterjs-server/blob/99368ede61e21fd7e1d798554d67735b92934495/packages/ws-controller/src/controller/ControllerCommandHandler.ts#L336-L339

```ts
const fields: Record<string, unknown> = {
    ...payload,
    originatingEndpointId,
};
```

Clients following the Python CHIP SDK wire convention (as generated in this very repo, see `python_client/chip/clusters/cluster_defs/WebRtcTransportProvider.py#L92`) send the field as `webRtcSessionID` (capital `ID`):

```py
ClusterObjectFieldDescriptor(Label="webRtcSessionID", Tag=0, Type=typing.Union[Nullable, uint]),
```

but matter.js's own generated property name for this field is `webRtcSessionId` (lowercase `d`), as used elsewhere in the same handler (`response.webRtcSessionId`).

Because the key doesn't match, the field is effectively dropped, and matter.js's own mandatory-field validation then reports it as missing — even when the client correctly sent `null` for a fresh offer (mandatory nullable field).

This is exactly the class of bug that `convertCommandDataToMatter` (`packages/ws-controller/src/server/Converters.ts`) already exists to solve, and which `ControllerCommandHandler.handleInvoke` uses for regular (generic) invoke commands — but `sendWebRtcProviderCommand` bypasses it entirely.

## Reproduction

Send `send_webrtc_provider_command` with:

```json
{
  "node_id": ...,
  "endpoint_id": ...,
  "command_name": "ProvideOffer",
  "payload": {
    "webRtcSessionID": null,
    "sdp": "<offer sdp>",
    "streamUsage": 1
  }
}
```

Expected: command succeeds and returns a `ProvideOfferResponse` with an assigned session id.
Actual: fails with `ValidationMandatoryFieldMissingError/128: Missing mandatory field webRtcSessionId`.

## Suggested fix

Route `payload` through `convertCommandDataToMatter` (using `ClusterMap[WebRtcTransportProvider.id]`) before merging in the server-injected `originatingEndpointId`, the same way `handleInvoke` already does for generic commands. Happy to open a PR with this change plus a regression test in `packages/matter-server/test/converter/ConvertersTest.ts` if useful.

lboue:
## Summary
- `sendWebRtcProviderCommand` spread the incoming `payload` directly into the matter.js command `fields`. Clients following this project's own Python client wire convention (`python_client/chip/clusters/cluster_defs`) send `webRtcSessionID` (capital `ID`), but matter.js's own property name is `webRtcSessionId` (lowercase `d`). The mismatched key was silently dropped, and matter.js's mandatory-field validation then reported it as missing (`ValidationMandatoryFieldMissingError/128: Missing mandatory field webRtcSessionId`).
- Route the payload through the same `convertCommandDataToMatter` model-based conversion already used by `handleInvoke` for generic commands, so field names and value types (nullables, etc.) line up with what matter.js expects.
- The response side has the same problem in reverse: matter.js returns `webRtcSessionId`/`videoStreamId`/`audioStreamId` (its own camelCase), but HA's camera integration reads `response["webRtcSessionID"]`. Added `convertWebRtcProviderResponseToWebSocket` to rename the known response fields back to the wire convention.
- Found and fixed the same acronym-casing bug for `CameraAvStreamManagement`'s `VideoStreamAllocate`/`AudioStreamAllocate` responses (`videoStreamId`/`audioStreamId` -> `videoStreamID`/`audioStreamID`), hit via the generic `device_command` invoke path. HA's Matter camera needs these two commands to allocate a video/audio stream before it can reference one in `ProvideOffer`. Added `convertCameraStreamAllocateResponseToWebSocket`, wired into `#convertCommandDataToWebSocket` for cluster `0x0551` alongside the existing WebRTC provider special-case.
- Added regression tests for all of the above.

Fixes #812

## Test plan
- [x] `npx tsc --noEmit -p packages/ws-controller/src/tsconfig.json` passes with no new errors
- [x] `Converters` test suite passes (103/103)
- [x] `npm run format-verify` passes


Apollon77:
Closing in favor of #824, which adopts the request-side fix from this PR (routing the payload through `convertCommandDataToMatter`) and generalizes the response side: the CHIP wire-name conversion (`ACRONYMS`/`FIELD_NAME_OVERRIDES`) is extracted from the python client generator into a shared module and applied in `convertMatterToWebSocketNameBased`, so all command responses and event data emit wire names (`webRtcSessionID`, `videoStreamID`, `groupID`, ...) without per-command rename maps. Thanks @lboue for the excellent analysis and groundwork!

PR #826, opened 2026-07-08
feat: add get_fabric_label WS command and --default-fabric-label pin

Apollon77:
## Problem

Two Home Assistant instances pointed at one Matter server could each set a different fabric label via `set_default_fabric_label`, playing name ping-pong.

## Changes

- **New WS command `get_fabric_label`** (schema 12) — returns the current fabric label so clients read it instead of assuming their own value. Read live from matter.js, falling back to the configured label.
- **New CLI flag `--default-fabric-label`** (env `DEFAULT_FABRIC_LABEL`) — pins the label: overrides any persisted/legacy value, reaches matter.js on both fresh and existing fabrics, and makes `set_default_fabric_label` a logged no-op while pinned.
- **Clients** — ws-client (`getFabricLabel`) and python client (`get_fabric_label`), both `require_schema` 12.
- **Dashboard** — Fabric Label row in server-details (schema-12 guarded) with a pencil-edit dialog that calls `set_default_fabric_label` and re-fetches on save. Per design decision, no lock-aware UI: when pinned the set is silently ignored server-side (logged).
- **Docs** — `cli.md`, `websockets_api.md`, WS schema changelog.
- **Tests** — ConfigStorage lock unit test; integration tests now verify the label via `get_fabric_label`.

## Verification

`npm run format` / `lint` / `build` clean; full `npm test` 251/251 (incl. live integration). Per-change and a final cumulative adversarial review run — no bugs found. Dashboard verified by build + tests, not yet driven in a live browser.


Apollon77:
Thanks @copilot. Addressed in c4ce6b6:

1. **ConfigStorage lock-before-persist** — fixed. `lockFabricLabel` now persists the label *before* setting `#fabricLabelLocked`, so a failed `set()` can't leave the instance locked with a stale label.

3. **Dashboard offline alert / `err.message`** — fixed. `#refreshFabricLabel` now skips the fetch when `connection.connected` is false and logs failures via `console.error` instead of a modal alert (the modal is kept only for explicit user actions like opening the editor).

2. **Pin reaching matter.js** — declining; already handled. The pinned value is written to `ConfigStorage.fabricLabel` before `MatterController.create()`, and the controller passes it as `adminFabricLabel`. matter.js reconciles an existing fabric on start — `FabricAuthority.defaultFabric` does `if (fabric.label !== config.adminFabricLabel) await fabric.setLabel(...)`, and `CommissioningController.validateAndUpdateFabricLabel` pushes it to connected nodes. So `get_fabric_label` (live value, with config fallback while the controller is still starting) returns the pinned label on both fresh and existing fabrics. An extra post-start `updateFabricLabel` would be a redundant write.

Apollon77:
Thanks @copilot — fixed in 60d767b. `sendOn` now registers the error handler with `on` (not `once`) and removes both the message and error listeners via a shared `cleanup()` when the request settles, so repeated calls on a long-lived connection no longer accumulate listeners.

Apollon77:
Addressed in 5637714:

- **server-details fabric-label refresh**: removed the offline early-return. The component only mounts under the connected view (and re-mounts on reconnect, re-running `firstUpdated`), so the guard was dead code that read as an early-return without a re-fetch hook. Failures now just log via `console.error` — no stuck placeholder in practice.
- **CHANGELOG**: fixed `WebSocket` capitalization, "other connects" → "other connections", and dropped the `fabric-label` hyphen.

Apollon77:
Addressed in 3b6ba10:

- **`set_default_fabric_label` write order**: now applies to matter.js first and persists to `ConfigStorage` only on success, so a failed controller update no longer desyncs the stored label from the live fabric label.
- **`fabric-label-dialog` input cast**: reads `.value` via `MdOutlinedTextField` instead of the incorrect `HTMLInputElement` cast.
- **`sendOn` timeout**: declining — mocha's per-test timeout already fails a hung test, and the sibling `handle` helper has no timeout either; a shorter timer would just duplicate that.

PR #832, opened 2026-07-09
feat(ws-controller): network topology service + get_network_topology API (schema 13)

### LINKED ISSUE #210: Add Matter network visualization in Matter panel and aim for a consistent graph experience across protocols
### Problem statement

### Community signals

#### In scope

#### Not in scope

### Foreseen solution

### Risks & open questions

### Appetite

Medium - 3-4 weeks

### Execution issues

_No response_

### Decision log

| Date | Decision | Outcome |
|------|----------|---------|
|      |          |         |

MindFreeze:
Part of the Matter network visualization work for [OpenHomeFoundation roadmap #210](https://github.com/OpenHomeFoundation/roadmap/issues/210). Follows #829 (the shared topology-derivation extraction), which this builds on.

Derives the full Matter network graph (Thread mesh + Wi-Fi star) server-side and exposes it over the WebSocket API, so downstream consumers (Home Assistant core / frontend) can relay a ready-made topology instead of each re-deriving it from raw cluster attributes and lacking the server-only mDNS Border Router data.

## What's here

- **`NetworkTopologyService`** — builds the graph from the shared `ws-client` derivation pipeline plus the passively-discovered Border Router registry. Conventions mirror `ThreadDiagnosticsService` (injected `Pick<>` deps, an `Observable` event, testable without a controller). Hybrid update model: debounced change-driven rebuilds, emit-only-when-changed (hash compare excluding the timestamp), and a slow periodic rebuild for sleepy-device drift. `refresh()` re-reads Thread neighbor/route tables (and Wi-Fi diagnostics) from online nodes, concurrency-capped under an overall deadline.
- **`get_network_topology { refresh? }`** command + **`network_topology_updated`** event, bumping `SCHEMA_VERSION` to **13** (min supported stays 11). The event is delivered only to connections that have issued the command, mirroring the schema-12 `thread_diagnostics_updated` opt-in so pre-13 clients never receive an event type they didn't subscribe to.
- **Wire types** (`NetworkTopology` / node / connection, snake_case): both observed Thread link directions are preserved, with a strongest-direction summary `strength` for simple renderers.

## Scope

This first iteration derives Thread links from the nodes' own `ThreadNetworkDiagnostics` (neighbor + route tables) and classifies externals against the BR registry. The richer MeshCoP diagnostic enrichment (route64 / childTable → router-to-router links and diagnostic-only mesh nodes) is a planned follow-up; the wire model already accommodates it.


MindFreeze:
I'll keep this in draft until I have tested the whole stack. May need to tweak the API

packages/ws-controller/src/controller/MatterController.ts
```diff
@@ -457,6 +459,30 @@ export class MatterController {
...
+     * Lazily-constructed network topology service. Created on first access (i.e. the first
+     * client that queries/subscribes topology), wired to the command handler's node cache +
+     * events and the Border Router registry. Reused across connections.
+     */
+    get networkTopology(): NetworkTopologyService {
+        if (this.#networkTopology === undefined) {
+            const handler = this.commandHandler;
+            this.#networkTopology = new NetworkTopologyService({
+                listNodes: () => handler.getNodeIds().map(nodeId => handler.getNodeDetails(nodeId)),
+                readAttributes: (nodeId, paths) =>
```
Apollon77: **Blocker: this makes `get_network_topology { refresh: true }` a no-op** (slower identical snapshot + real radio traffic to every node).

`handleReadAttributes(nodeId, paths, fabricFiltered = false)` — the third arg defaults to `false`. matter.js `ClientStructure.#mutateAttribute` has a filter-*match* gate:

```js
if (this.subscribedFabricFiltered !== scope.isFabricFiltered) return currentUpdates;
```

`subscribedFabricFiltered` defaults to `true`, so a `fabricFiltered=false` read is dropped for state integration — no store write, no `PairedNode.events.attributeChanged`, therefore `attributeCache.updateAttribute(...)` (`ControllerCommandHandler.ts:576`) never runs. Since the read result is also discarded here (`.then(() => undefined)`), the fresh values go nowhere and `#build()` re-reads the unchanged cache. The gate is a pure filter match, so it bites even for non-fabric-scoped clusters like Thread Diagnostics (53).

The dashboard already hit exactly this and fixed it — `packages/dashboard/src/pages/network/update-connections-dialog.ts:188`:

> `// fabric_filtered must match the node subscription's filter (true), else matter.js discards the read and no attribute_updated fires, leaving the store stale.`

Fix:

```ts
readAttributes: (nodeId, paths) =>
    handler.handleReadAttributes(NodeId(nodeId), paths, true).then(() => undefined),
```

(`includeKnownVersions: true` is already set in `handleReadAttributes`, which is required to keep dataVersion filters off once `fabricFiltered` is `true` — so no second change needed.)

packages/ws-controller/src/controller/MatterController.ts
```diff
@@ -457,6 +459,30 @@ export class MatterController {
         return this.#threadDiagnostics;
     }
 
+    /**
+     * Lazily-constructed network topology service. Created on first access (i.e. the first
+     * client that queries/subscribes topology), wired to the command handler's node cache +
+     * events and the Border Router registry. Reused across connections.
+     */
+    get networkTopology(): NetworkTopologyService {
```
Apollon77, line 482: The lazy getter has no `#stopped` guard. A `get_network_topology` that lands after `stop()` constructs a **fresh** service — 60s `setInterval` plus subscriptions to the controller/BR observables — which nothing will ever stop, since `stop()` already ran and only stops `#networkTopology` once. The handler's `#shuttingDown` flag guards sends, not the command path.

Suggest returning the existing instance only, or checking `#stopped` before constructing.

packages/ws-controller/src/controller/MatterController.ts
```diff
@@ -457,6 +459,30 @@ export class MatterController {
...
+        if (this.#networkTopology === undefined) {
+            const handler = this.commandHandler;
+            this.#networkTopology = new NetworkTopologyService({
+                listNodes: () => handler.getNodeIds().map(nodeId => handler.getNodeDetails(nodeId)),
+                readAttributes: (nodeId, paths) =>
+                    handler.handleReadAttributes(NodeId(nodeId), paths).then(() => undefined),
+                borderRouters: this.#borderRouterRegistry,
```
Apollon77, line 491: With `disableThreadDiagnostics`, this registry is constructed but never started, so `list()` returns `[]` for the process lifetime and every unmatched neighbour falls through to `thread_unknown`. Not wrong, but the payload gives a consumer no way to distinguish "no border routers discovered" from "BR discovery is switched off".

packages/ws-controller/src/controller/NetworkTopologyService.ts
```diff
@@ -0,0 +1,505 @@
...
+
+    /**
+     * Re-read the Thread neighbor/route tables (and WiFi diagnostics) from every online
+     * controller node (aux-source nodes are never read), then rebuild. Reads are
+     * concurrency-capped and best-effort (a failing/slow node is skipped), mirroring the
+     * dashboard's "update connections" action. Once the overall deadline expires no further
+     * reads start; reads already in flight run to completion (their results land in the
+     * cache and surface via the next rebuild).
+     */
+    async refresh(): Promise<NetworkTopology> {
```
Apollon77, line 196: No in-flight coalescing: the concurrency cap is per call, so N concurrent `get_network_topology { refresh: true }` callers (HA core + dashboard + a second HA instance) each fan out reads to every online node — N× the radio traffic, and each caller's deadline runs independently.

`ThreadDiagnosticsService` solves the same problem with `#streamsInFlight` keyed dedupe in `getOrFetch`. Suggest the same here: keep a single in-flight `Promise<NetworkTopology>` and have concurrent callers await it (optionally plus a short result TTL so back-to-back refreshes are cheap).

packages/ws-controller/src/controller/NetworkTopologyService.ts
```diff
@@ -0,0 +1,505 @@
...
+        network_name: ext.networkName,
+    };
+}
+
+/**
+ * Fold the 0-2 directional edges of a Thread pair into one wire connection. Both directions are
+ * preserved (asymmetry stays legible); the summary `strength` is the strongest observed direction.
+ * Pairs where every observed direction is dead (LQI 0 → "none") are dropped.
+ */
+function mapThreadPair(pair: ThreadEdgePair): NetworkTopologyConnection | undefined {
```
Apollon77, line 457: Dropping all-`none` pairs here leaves orphan nodes in the graph. `findUnknownDevices` materializes an external for *every* unmatched neighbour, including LQI 0 entries (it doesn't filter on link quality), and `getSignalLevelFromLqi` maps `lqi <= 0` to `"none"`. So a node whose only neighbour record is a stale/dead entry ships as a `thread_unknown` (or `border_router`) node with zero connections — a floating dot for every renderer.

Either keep the edge at `strength: "none"` (renderers can hide it) or prune externals that end up with no edges after the pair loop.

packages/ws-controller/src/controller/NetworkTopologyService.ts
```diff
@@ -0,0 +1,505 @@
...
+ * An attribute change outside this set can't affect the graph, so it doesn't schedule a rebuild.
+ */
+const TOPOLOGY_CLUSTERS = new Set<number>([49, 51, 53, 54]);
+
+/**
+ * Thread neighbor + route tables, interface list, own RLOC16 — the inputs to Thread edge
+ * derivation. RLOC16 (0/53/64) is included because edge/unknown matching falls back to it
+ * when extended-address matching fails, and it changes when a device re-attaches to the mesh.
+ */
+export const THREAD_REFRESH_PATHS = ["0/53/7", "0/53/8", "0/51/0", "0/53/64"];
```
Apollon77: `0/53/1` (RoutingRole) and `0/53/2` (NetworkName) are missing, but both feed the built node (`role`, and see the `network_name` comment below). A device that re-attaches as a REED/router keeps its stale `role` through an explicit `refresh`, which is the one code path a user triggers *because* they want current data.

packages/ws-controller/src/controller/NetworkTopologyService.ts
```diff
@@ -0,0 +1,505 @@
...
+                    kind: "matter",
+                    network_type: "thread",
+                    node_id: node.node_id,
+                    role: mapThreadRole(getThreadRole(node)),
+                    available,
+                    is_bridge: isBridge,
+                    ext_address: getThreadExtendedAddressHex(node),
+                    rloc16: getThreadRloc16(node),
+                    ext_pan_id: extPanIdHex,
+                    network_name: extPanIdHex !== undefined ? networkNameByXp.get(extPanIdHex) : undefined,
```
Apollon77: `network_name` for commissioned Thread nodes resolves only via the BR registry (extPanId → name). With no BR discovered — or with BR discovery disabled — it stays `undefined` even though the node itself reports NetworkName in `0/53/2`. Suggest falling back to the node's own attribute (and adding `0/53/2` to `THREAD_REFRESH_PATHS`).

packages/ws-controller/src/controller/NetworkTopologyService.ts
```diff
@@ -0,0 +1,505 @@
...
+}
+
+const LEVEL_RANK: Record<TopologyStrength, number> = { none: 0, weak: 1, medium: 2, strong: 3 };
+
+function strongestLevel(levels: TopologyStrength[]): TopologyStrength {
+    return levels.reduce((best, level) => (LEVEL_RANK[level] > LEVEL_RANK[best] ? level : best), "none");
+}
+
+/** WiFi RSSI → strength (ports the dashboard's -70/-85 dBm thresholds). Unknown RSSI is treated as medium. */
+function rssiToStrength(rssi: number | null): TopologyStrength {
```
Apollon77, line 488: Nit: an unknown RSSI reporting as `"medium"` is indistinguishable on the wire from a genuinely medium link, and `strength` is a required field, so consumers can't detect the synthetic value. Consider `"none"`, or make the direction info absent when there is no measurement.

packages/ws-controller/test/NetworkTopologyServiceTest.ts
```diff
@@ -0,0 +1,655 @@
...
+                mkThread(3, { role: 5, rloc16: 1026, available: false }),
+                mkEthernet(4),
+            ];
+            const { service, readCalls } = makeHarness({ nodes: () => nodes });
+
+            const topology = await service.refresh();
+
+            expect(topology.nodes.length).to.be.greaterThan(0);
+            const byNode = new Map(readCalls.map(c => [c.nodeId, c.paths]));
+            expect(byNode.get(1)).to.deep.equal(THREAD_REFRESH_PATHS);
```
Apollon77, line 573: This is where the `fabricFiltered` blocker slips through: the harness stubs `readAttributes` entirely, so the test proves the right *paths* reach the right nodes but nothing about how the read is issued or whether its result ever lands in the cache. Worth asserting the `fabricFiltered` argument at the `MatterController` wiring level (or covering the full read → cache → rebuild path once), otherwise the same regression can return silently.

Apollon77 (commented):
Review pass. Solid shape overall — the shared `ws-client` derivation reuse, the lazy per-connection observer, the sorted hash, and the stop-mid-refresh handling are all right. One functional blocker plus a few robustness items, inline.

**Blocker:** `refresh` never actually refreshes — the attribute read is issued with `fabricFiltered=false`, which matter.js discards for state integration, so the rebuild sees the same stale cache. Details on `MatterController.ts`.

Not inline-able, one more item: with `--disable-thread-diagnostics` the `BorderRouterRegistry` is constructed but never started, so `list()` stays empty and *every* Thread neighbour classifies as `thread_unknown` — the graph silently loses all `border_router` nodes. Worth a line in the schema changelog (or a flag in the payload) so consumers don't read the absence as "no BRs on the network".

packages/ws-controller/src/controller/MatterController.ts
Apollon77: Fixed in e43d2c8d: the read now goes through an exported `topologyAttributeReader(handler)` that passes `fabricFiltered=true`, with the matter.js filter-match gate recorded on it. Leaving this thread open for you to review the fix.

packages/ws-controller/src/controller/MatterController.ts
Apollon77: Fixed in e43d2c8d: the getter now throws `Controller is stopped` instead of constructing a service after shutdown, and `ensureTopologyObserver` bails when the connection is closed/shutting down so the throw cannot swallow a request reply. Leaving open for review.

packages/ws-controller/src/controller/MatterController.ts
Apollon77: Documented in e43d2c8d (schema-13 changelog entry) rather than changed in the payload: with Thread diagnostics disabled, no `border_router` nodes appear and every external neighbour reports as `thread_unknown`. Leaving open in case you would rather surface it in the wire model.

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Fixed in e43d2c8d: `refresh()` keeps a single in-flight promise (`#refreshInFlight` + `#runRefresh`), so concurrent callers share one read fan-out and a later call starts a fresh run. Covered by a new test. Leaving open for review.

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Fixed in e43d2c8d: `#build()` now builds the edges first and materializes an external only if some connection references its id, so a dead-link-only (LQI 0) neighbour no longer ships as an edgeless node. Leaving open for review.

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Fixed in e43d2c8d: `0/53/1` and `0/53/2` added to `THREAD_REFRESH_PATHS`. Leaving open for review.

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Fixed in e43d2c8d: `network_name` now prefers the node's own NetworkName via a new shared `getThreadNetworkName()` in `ws-client`, falling back to the Border Router registry lookup. Leaving open for review.

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Not changed — left as your deliberate parity with the dashboard thresholds. Flagging only so the synthetic value is a known property of the wire contract. Leaving open for your call.

packages/ws-controller/test/NetworkTopologyServiceTest.ts
Apollon77: Addressed in e43d2c8d: added a `controller wiring` test asserting `topologyAttributeReader` issues the read with `fabricFiltered=true`, plus tests for refresh coalescing, the dropped dead-link external, and the network-name preference. Leaving open for review.

packages/ws-controller/src/controller/NetworkTopologyService.ts
MindFreeze: I'll change it to `none`

packages/ws-controller/src/controller/MatterController.ts
MindFreeze: I think we can keep it just documented

packages/ws-controller/src/controller/NetworkTopologyService.ts
Apollon77: Following up on 7596196b: mapping an unmeasured RSSI to `"none"` fixes the undetectable-synthetic-value problem but lands on the one value that already means something else. `mapThreadPair` drops Thread pairs whose every observed direction is `"none"`, so a Thread edge can never *be* `"none"` on the wire, and consumers use exactly that convention as a liveness test — the dashboard filters on `signalLevel !== "none"` in `network-utils.ts:488`, `thread-graph.ts:466-467`, `:576` and `:655`. A Wi-Fi station whose RSSI cannot be read would ship an edge that any such consumer hides, leaving a live station rendered unlinked from its AP.

Proposed fix, prepared as a commit: `TopologyStrength` gains `"unknown"` ("no measurement, link presumed up"), `rssiToStrength(null)` returns it, and the changelog + `TopologyStrength` doc spell out the `none` vs `unknown` split. Your `source_to_target: undefined` for the unmeasured case is kept as-is. Schema 13 is unreleased so the union is free to grow, and #917 models `strength` as an open string set, so the Python client parses it unchanged.

Dashboard visualisation is unaffected: it never imports the wire `NetworkTopology`/`TopologyStrength` types — it derives its own graph and colours Wi-Fi edges from raw RSSI via `getSignalColorFromRssi` (`null` → medium colour). Worth aligning when the dashboard switches to consuming the server-derived graph: `unknown` should render neutral-but-live there rather than reusing the medium colour.

Gates on the prepared commit: format/lint/build clean, ws-controller 353/353, ws-client 79/79, dashboard 153/153.

Apollon77: Pushed as b3a91ed8.

Issue #843, opened 2026-07-10
Subscription can die silently (no timeout, no re-subscribe, node stays available) — forensic timeline on 1.1.7, wedge-mode reproduced on 1.2.2


mtflud:
## Summary

A node's active subscription can stop delivering reports while the server's client-side
subscription timeout never fires. The node remains `available: true` with frozen attribute
state indefinitely (observed: 14+ hours), while other controllers on other fabrics (Apple
Home) keep receiving updates. Availability at the ws-controller layer is purely
connection-state driven, so nothing detects the dead subscription.

## Environment

- HAOS add-on (matter-server 1.1.7 / matter.js 0.17.4) for the original incident
- Reproduced wedge-mode on matter-server 1.2.2 / matter.js 0.17.5-alpha.0-20260710 (source build)
- 9 commissioned Thread nodes (Eve Door/Energy/Motion contact+plug devices, Onvis plugs),
  3 fabrics on the devices (two Apple, one HA)

## Forensic timeline (incident, 1.1.7)

Day 1 had several server restarts in close succession (host VM migration), then:

- 18:35 first boot: subscription `3ab415cf` to node 0x12 (Eve Energy plug, negotiated
  interval 1m6s / timeout 1m44s)
- 18:58 second boot: new subscription `cb5158b0` (interval 1m1s / timeout 1m39s). The
  node's HA-visible state froze at exactly this moment.
- 22:25 node 0x12 delivers data reports on its *current secure session* for FOUR
  subscription IDs unknown to the server — including `3ab415cf` from the 18:35 boot —
  all logged `Ignoring data report for unknown subscription ID` and discarded. Another
  node (0x13) did the same at 23:29.
- For the following 14+ hours: no reports processed for 0x12, **no subscription timeout
  fired** (the only timeout+re-subscribe in the whole log window was a different node's),
  no availability change, state frozen. Recovery required an add-on restart.

## Reproduction on 1.2.2 (source build + external liveness watchdog)

After 3 add-on restarts ~30s apart (9 nodes re-subscribe cleanly in <20s):

- node 0x12: subscription established 14:20:10 (interval 1m4s / timeout 1m42s); matter.js
  itself detected one loss at 14:22:15 ("timed out after 1m42s" → replace → OK); the
  replacement subscription then went silent and **no timeout fired** — an external
  watchdog we added at the ws-controller layer (stamping `PairedNode.events.connectionAlive`)
[...]

PR #844, opened 2026-07-10
fix(ws-controller): five robustness fixes — guarded observer/build/cache throws, observer-group cleanup, time-sync cooldown, unhandled-rejection hardening

mtflud:
Companion to #843 (frozen-state forensics). Five independent fixes, one commit each, every one with a regression test (or a stated-reason exemption in the commit body):

1. **Guard direct-mode frame `build()` + `attributeChanged`/`eventChanged` observers** — a device-data-dependent converter throw (`Invalid bitmap value`, `BigInt` RangeError) escaped synchronously through `Observable.emit` into PairedNode's report dispatch, aborting delivery of the remaining attributes in the batch and starving other connections' observers (mode-dependent, intermittent frozen states). Includes a differential two-connection regression test proving one connection's throw no longer starves the other in the same emit.
2. **Guard `AttributeDataCache.updateAttribute`** — same throw family in the cache layer; the identical conversion in `#collectAttributes` was already guarded. (Note: the guard catches plain `Error` too — `Invalid bitmap value` is not a `MatterError`, so `MatterError.accept` alone would rethrow it.)
3. **Close the stale per-node ObserverGroup before replacing it in `#registerNode`** — re-registration would double-register every node observer.
4. **Time-sync trigger cooldown only on success** — the 24 h cooldown was stamped *before* the attempt, so one transient failure locked out reactive re-sync for a day; failures now get a 5-minute retry backoff instead.
5. **`.then(onFulfilled, onRejected)` → `.then().catch()`** in the WS response path — a throw during response serialization became an unhandled rejection (process exit on modern Node); regression test captures the real `unhandledRejection` pre-fix.

All commits cherry-pick cleanly onto current `main` (25f3097); suite green (239/239). Happy to split into separate PRs if you prefer.

https://claude.ai/code/session_019cYKQfvTvuyjeWR2q73VJy

### DIFF
--- a/packages/ws-controller/src/controller/AttributeDataCache.ts
+++ b/packages/ws-controller/src/controller/AttributeDataCache.ts
@@ -73,17 +73,39 @@ export class AttributeDataCache {
      */
     updateAttribute(nodeId: NodeId, data: DecodedAttributeReportValue<any>): void {
         const { endpointId, clusterId, attributeId } = data.path;
+        const path = buildAttributePath(endpointId, clusterId, attributeId);
 
         const clusterData = ClusterMap[clusterId];
-        const convertedValue = convertMatterToWebSocketTagBased(
-            data.value,
-            clusterData?.attributes[attributeId],
-            clusterData?.model,
-        );
+        let convertedValue: unknown;
+        try {
+            convertedValue = convertMatterToWebSocketTagBased(
+                data.value,
+                clusterData?.attributes[attributeId],
+                clusterData?.model,
+            );
+        } catch (error) {
+            // Unlike #collectAttributes below (which runs inside #runPopulate's own catch-all), this
+            // path has no safety net: it is invoked directly from the node's attributeChanged
+            // observer, so a rethrow here would escape into the event dispatch and could abort
+            // delivery of every other pending update for this node. A single malformed value must
+            // never take that down, so always log and skip — regardless of error type, unlike the
+            // MatterError-only tolerance below.
+            if (error instanceof MatterError) {
+                logger.debug(
+                    `Ignoring attribute update ${path} for node ${formatNodeId(nodeId)} because of`,
+                    Diagnostic.errorMessage(error),
+                );
+            } else {
+                logger.warn(
+                    `Ignoring attribute update ${path} for node ${formatNodeId(nodeId)}: converter threw`,
+                    error,
+                );
+            }
+            return;
+        }
         if (convertedValue === undefined) {
             return;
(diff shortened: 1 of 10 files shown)


Apollon77:
Hm ... >800k of HA users use the current version and effectively all is tested in >4 month of Beta and no crashes were reported. So before adding try catches around I want to see logs where this actually happens. It is not needed to add "hardening" without the need to add hardening.

So please provide logs and details about concrete issues and then we can discuss how to best fix them. I leave this PR open until I see this. If not we close it.

Apollon77:
I had a look at all the commits:

* guard direct-mode coalescable build and node-event observers
  * The code is normally written in a way that there shouldbe no cases that throw, so befroe addign plain try-catches in "theoretically culd thor" places I would like to see a place where it really throws and then better fix the reason than adding "Highest level guards". I like to address root causes and not symthoms
* guard AttributeDataCache.updateAttribute against a rejected value
  * Same as above. In all normal code flows only allowed values come in here and are parsed as by definition and model and in fact matter,js should ensure that this is all "right". So when value conversion would throw I want to understand the root cause and better fix this because the consequences can be much more than that and it is not really a solution to not emit such a changes value because in fact we miss a value change then. Also here: Fixing root causes and not sympthoms
* close the stale per-node ObserverGroup before replacing it
  * I do not see the praxis relevance. Recommission with the same node id as an existing one will now happen because declined by matter.js. #registerNode is exactly called in two places: on startup and when commissioning a new node afetr the commissioning was successful. There is no logic path that can call this aside this, so it is a theoretical case and honestly before fixing like that I would better throw here an error because this case must never happen, so when I would love to know to fix the root cause.
* only apply the 24h trigger-sync cooldown on success
  * This point is ok but code is intended like this. Time sync depends on the node nd usually the time is in sync or is added oif devcie reboots and looses time after the reconnect ... otherwise the assumptionis more that an "time update every24h" is just there for the sake of "keep time shift minimal"; so it is ok to just retry 24h later and no need to spend more traffic. If there is a practice issue with this approach please state this as issue and we can discuss how to solve this. For me, there is no need to add more complexity in here right now
* stop an unhandled rejection from a throw during response serialization
  * This one is basically correct when assuming that the json serialization would throw ... And then we are again at the case of "why ... find root cause and fix this one" rather than hiding it here, but I might consider to adopt this one
* give TimeNotAccepted the full trigger-sync cooldown
  * This one is valid and I will adopt it ... if it was existing before your other time sync changes but I think irrelevant

PR #845, opened 2026-07-10
feat(ws-controller): subscription-liveness watchdog (opt-in) — detects silently-dead subscriptions and forces resubscribe

mtflud:
Implements the recovery layer proposed in #843. A `SubscriptionWatchdog extends NodeProcessor` that:

- stamps every `PairedNode.events.connectionAlive` per node (data reports and keepalives);
- trips when a Connected node is silent past `1.5 × currentSubscriptionIntervalSeconds + 60 s` (15-min fallback when the interval is unknown, logged as suspicious);
- on trip: transition-gated `forceUnavailable` → `nodeAvailabilityChanged(false)` (no duplicate events on repeated trips), then `triggerReconnect()`; per-node backoff `max(2×threshold, 10 min)`, cleared by any fresh `connectionAlive`;
- after the post-trip reconnect, forces the attribute-cache rebuild and a full `node_updated` broadcast even on the fast-reconnect path (extracted into a shared `runConnectedCacheRepair` helper so the ordering/gating contract is pinned by tests);
- hourly per-node health-snapshot log line; all error paths contained per the NodeProcessor contract.

**Defaults OFF upstream** (`--subscription-watchdog` / `SUBSCRIPTION_WATCHDOG`), so behavior is unchanged unless opted in.

Field evidence (same environment as #843, watchdog enabled): after 3 rapid server restarts, two Thread plug nodes' fresh subscriptions died silently — matter.js's own timeout recovered one earlier loss but missed these — and the watchdog detected them at 153 s and 160 s of silence respectively and fully recovered both (probe read + new subscription + availability cycle). Logs in the issue.

16 unit tests (silence/trip/backoff/reset/fallback/error-containment/cache-repair contract); suite green on current `main` (246/246 with this branch).

https://claude.ai/code/session_019cYKQfvTvuyjeWR2q73VJy

### DIFF
--- a/packages/matter-server/src/MatterServer.ts
+++ b/packages/matter-server/src/MatterServer.ts
@@ -188,6 +188,7 @@ async function start() {
             serverVersion: MATTER_SERVER_VERSION,
             bleProxyEnabled: cliOptions.bleProxy,
             enableTimeSync: cliOptions.enableTimeSync,
+            subscriptionWatchdog: cliOptions.subscriptionWatchdog,
             disableThreadDiagnostics: cliOptions.disableThreadDiagnostics,
         },
         legacyServerData,
--- a/packages/matter-server/src/cli.ts
+++ b/packages/matter-server/src/cli.ts
@@ -221,7 +221,7 @@ export function parseCliArgs(argv?: string[]): CliOptions {
             )
                 .argParser(parseBooleanEnv)
                 .preset(true)
-                .default(true)
+                .default(false)
                 .env("SUBSCRIPTION_WATCHDOG"),
         )
         .addOption(
--- a/packages/matter-server/src/cli.ts
+++ b/packages/matter-server/src/cli.ts
@@ -70,6 +70,9 @@ export interface CliOptions {
     // Time synchronization configuration
     enableTimeSync: boolean;
 
+    // Subscription liveness watchdog configuration
+    subscriptionWatchdog: boolean;
+
     // Dashboard configuration
     disableDashboard: boolean;
     productionMode: boolean;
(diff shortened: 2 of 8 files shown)


mtflud:
Heads-up on a liveness caveat discovered in extended testing, plus two follow-up commits on this branch.

The watchdog's liveness source (`PairedNode.events.connectionAlive`) does not currently receive empty subscription keepalives — see matter-js/matter.js#4057. Until that lands, enabling the watchdog produces predictable false positives on healthy quiet devices: they get flagged right at the detection threshold (`1.5 × maxInterval + 60s`) while their empty keepalives are arriving on schedule, then force-reconnected. The default-off/opt-in shape of this PR is therefore load-bearing, not just cautious.

New commits since the PR was opened:

- a clamp treating implausible reported intervals (> 24h — e.g. the `Hours.one`-milliseconds fallback of `SustainedSubscription.maxInterval`, see matter-js/matter.js#4058) as "interval unknown", so the watchdog falls back to its 15-minute threshold instead of silently disabling itself for the node;
- a doc-comment on `recordAlive` making the keepalive dependency explicit.

In our fork deployment we run the watchdog together with a `@matter/node` dist patch that emits `subscriptionAlive` from `keepaliveReceived` (proposed in matter-js/matter.js#4057), plus an end-to-end integration test that idles a commissioned node past the threshold and asserts empty keepalives keep it clear of false trips.

Apollon77:
Sorry but no. This is the wrong place to fix it. IF there is really an issue (not proofen, logs requested) then the fix needs to go into matter.js and not here. Thats why I directly close this PR because. We will not work around potential issues in matter.js. we will fix them there

PR #847, opened 2026-07-11
fix: concurrent camera snapshot + non-blocking startup init + reinterview navigation

Apollon77:
## Summary

Three fixes discovered while debugging camera snapshots on an encoder-constrained camera, verified live against the device.

### 1. Camera snapshot while a live stream is running (`feat`)
On cameras with `MaxConcurrentEncoders = 1`, the live WebRTC video stream holds the sole encoder, so `SnapshotStreamAllocate` returned **ResourceExhausted (0x89)** whenever it mapped to an encoder-requiring capability during streaming.

- Snapshot capability selection is now **encoder-aware**: while a video stream is allocated, prefer a `SnapshotCapabilities` entry with `RequiresEncodedPixels = false` and clamp the requested resolution to it; when idle, use the highest-resolution (encoder) capability.
- On such a camera this means a **concurrent** snapshot uses the encoder-free resolution and an **idle** snapshot uses the full resolution — the hardware limit of a single encoder.
- New `camera-stream-budget.ts` helper: derive video `maxFrameRate` from the camera's `MaxEncodedPixelRate` so a concurrent snapshot fits the shared encoder budget on pixel-rate-limited cameras, and reject reuse of an over-budget allocated video stream. Video `maxFrameRate` capped at 60 (was 120, which reserved the entire 1080p budget).

### 2. Stay on the node page after a reinterview (`fix`)
`_reinterview()` redirected to the overview (`location.hash = "#"`) on success, bouncing the user off the node detail page. The node still exists and `node_updated` refreshes it in place, so the redirect is removed. `_remove()` still navigates home (correct).

### 3. Keep vendor / border-router init off the node-init path (`refactor`)
Vendor-info warmup (DCL fetch) and border-router discovery were awaited before node initialization, delaying node connection / WS availability by ~1s. Both now run off the critical path, tracked in `#backgroundInit` and given a **bounded** chance to settle in `stop()` so they neither dangle into teardown nor block shutdown on slow/unreachable network work. A `#stopped` guard prevents a late `started` handler from spinning up background work after teardown.

## Testing
- `npm run format` / `npm run lint` / `npm run build` clean.
- Full `npm test` green (incl. matter-server integration).
- New unit tests: `CameraStreamBudgetTest` (budget math) and `SnapshotCapabilitiesTest` (encoder-aware selection, null-safe parsing).
- Verified live against an encoder-constrained camera: concurrent snapshot now succeeds; reinterview keeps the node page; startup reaches node init without the vendor/BR wait.

## Review
Passed an independent adversarial review over the cumulative diff; findings addressed (unbounded shutdown await → bounded; snapshot-during-connecting predicate → `_videoStreamId !== null`; unknown-budget reuse over-rejection; start-after-stop guard). PR review comments addressed: reservation clamp, per-dimension clamp, null-safe capability parsing, and generic (vendor-neutral) comments.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ This page shows a detailed overview of the changes between versions without the 
 ## **WORK IN PROGRESS**
 
 - Enhancement: Thread diagnostics fetch a Border Router's dataset via REST when no credentials are stored, enabling the faster MeshCoP (CoAP) transport instead of the slower REST collection
+- Fix: Update WebRTC and Camera-related logic to respect available Pixelrate and Encoders of the camera device
+- Fix: Optimize startup behavior
 
 ## 1.2.2 (2026-07-10)
 
--- a/packages/dashboard/src/components/webrtc-stream-view.ts
+++ b/packages/dashboard/src/components/webrtc-stream-view.ts
@@ -118,11 +118,10 @@ export function parseSnapshotCapabilitiesFromList(list: unknown[], preferEncoder
         };
     });
     // When preferEncoderFree (a video stream is live), restrict to encoder-free capabilities
-    // (RequiresEncodedPixels=false): with MaxConcurrentEncoders=1 (Aqara G350) the video stream
-    // holds the sole encoder, so only an encoder-free snapshot can be captured concurrently —
-    // on the G350 that is the 640×480 entry, while 1080p requires the encoder. When idle, use the
-    // full set so an idle capture can use the higher-resolution (encoder) capability. Within the
-    // chosen set, take the highest resolution.
+    // (RequiresEncodedPixels=false): a camera with MaxConcurrentEncoders=1 holds its sole encoder
+    // for the video stream, so only an encoder-free capability can be captured concurrently (often
+    // a lower resolution). When idle, use the full set so a capture can use the higher-resolution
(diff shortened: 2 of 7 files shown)


Apollon77:
Addressed the Copilot review comment in e6bc6923: `_snapshotBudgetReservation()` now clamps the requested resolution to the chosen capability via a shared `clampToCapability` helper, so the reservation matches what `_ensureSnapshotStream()` actually requests and no longer over-reserves px/s. (In practice this only affected cameras with no encoder-free capability — on the G350 the encoder-free path returns a 0 reservation before resolution is used.)

Apollon77:
Addressed both review comments in a8dd393a:
- `clampToCapability` now clamps each dimension independently (`Math.min` per axis), so it never increases a dimension.
- `parseSnapshotCapabilitiesFromList` filters `!== null` (asObject returns null, not undefined), so non-object list entries are skipped instead of crashing on property access. Added a regression test covering null/non-object entries.

Apollon77:
Addressed in fa258a9d: removed vendor names and specific model numbers from all code comments and test fixtures, describing the behaviour as general device quirks instead. This also fixes the specific comment you flagged — the empty-`SnapshotCapabilities`-list quirk is now described generically ("some cameras advertise SNP but ship an empty list") rather than attributed to a model whose fixture actually has multiple entries. Scrubbed the PR description too.

PR #860, opened 2026-07-13
feat(dashboard): QR code + formatted setup code in share dialog

Apollon77:
## What

The **Share device** dialog previously showed the raw 11-digit manual pairing code (e.g. `Setup code: 34315849284`). This PR:

- Formats the manual code per Matter spec as `XXXX-XXX-XXXX` (e.g. `3431-584-9284`)
- Renders the `setup_qr_code` payload — already returned by the server (`setup_qr_code` in `CommissioningParameters`) but previously unused — as a scannable QR code

## How

- New util `packages/dashboard/src/util/pairing-code.ts`:
  - `formatManualPairingCode()` — groups the 11-digit short code
  - `renderPairingQrCodeSvg()` — generates an SVG QR via `qrcode-generator` (zero-dep, ~10KB, pure JS)
- `node-details.ts` share dialog now uses a Lit `TemplateResult` (already supported by `showAlertDialog`)
- QR drawn black-on-white in a padded box so it stays scannable in dark mode

No server/protocol changes — dashboard-only.

## Verification

- `npm run format` / `npm run lint` / `npm run build` all green
- Util logic verified: `34315849284` → `3431-584-9284`; QR produces valid SVG
- Independent adversarial review: no issues

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5154,6 +5154,13 @@
                 "undici-types": "~8.3.0"
             }
         },
+        "node_modules/@types/qrcode-generator": {
+            "version": "0.0.16",
+            "resolved": "https://registry.npmjs.org/@types/qrcode-generator/-/qrcode-generator-0.0.16.tgz",
+            "integrity": "sha512-i+wPpeHH64qJHUIz51+tPq5FRkgMxE9/uhDE5eSlj3qujIjHm3U1wHfy8nMD/m1VKaXHO/uDjHamPVz2Fr4IPA==",
+            "dev": true,
+            "license": "MIT"
+        },
         "node_modules/@types/qs": {
             "version": "6.15.1",
             "resolved": "https://registry.npmjs.org/@types/qs/-/qs-6.15.1.tgz",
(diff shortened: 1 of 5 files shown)


Apollon77:
Addressed all six review comments in 0a77676:

- **unsafeSVG → `` data URI** — QR now rendered via `data:image/svg+xml`, keeping SVG parsing out of the DOM; `unsafeSVG` import removed.
- **Redundant white background** — dropped the hardcoded `#fff` wrapper; `createSvgTag()` already emits a white background rect.
- **Docstrings** — corrected both (`formatManualPairingCode` fallback behavior, and the QR "white background" vs the previous "transparent" wording).
- **Tests** — added `PairingCodeTest.ts` covering 11-digit grouping, non-digit stripping, non-11-digit codes, no-digit passthrough, and that the QR helper returns an `<svg>` data URI. Also wired a `test` script into the dashboard package so its tests actually run (previously there was none, so the existing `test/` suite wasn't executed).

All 76 dashboard tests pass; format/lint/build green.

PR #890, opened 2026-07-17
Document time sync option

saddfox:
Added `--enable-time-sync` to cli docs.
Added `ENABLE_TIME_SYNC` and `TZ` to docker docs.


Apollon77:
@saddfox Uups we really forgot that. Thanks a lot!

PR #917, opened 2026-07-23
feat(python-client): get_network_topology + network_topology_updated (schema 13)

### LINKED ISSUE #210: Add Matter network visualization in Matter panel and aim for a consistent graph experience across protocols

MindFreeze:
Python client counterpart of #832, for the Matter network visualization work in [OpenHomeFoundation roadmap #210](https://github.com/OpenHomeFoundation/roadmap/issues/210). Targets the next `matter-python-client` release (0.8.0), which Home Assistant core will consume.

## What's here

- **Typed wire models** — `NetworkTopology` / `NetworkTopologyNode` / `NetworkTopologyConnection` / `TopologyDirectionInfo` dataclasses matching the schema-13 wire contract. `kind` / `role` / `strength` are open string sets (not enums) so a newer server introducing values can never break `dataclass_from_dict` parsing.
- **`get_network_topology(refresh=False)`** — gated with `require_schema=13` (raises `ServerVersionTooOld` against older servers); issuing it also opts the connection in to topology pushes, per the server contract.
- **Typed event parsing** — `network_topology_updated` reaches subscribers as a parsed `NetworkTopology` (`EventType.NETWORK_TOPOLOGY_UPDATED`).

The client baseline `SCHEMA_VERSION` stays 11 — per-command gating is the feature gate, so the client keeps working against schema-11/12 servers.

## Testing

`ruff` / `mypy` clean; unit tests cover command translation + `refresh` passthrough, the schema-12 rejection path, forward-compat parsing (unknown fields and enum-like values), and typed event dispatch. The full python integration suite stays green.

Also verified live against a server build from #832: connect (schema 13) → empty typed graph → `import_test_node` fixtures → `network_topology_updated` push parsed into `NetworkTopology` (18 nodes / 16 connections with per-direction LQI/RSSI) → `refresh=True` round-trip.


python_client/matter_server/client/client.py
```diff
@@ -308,6 +309,22 @@ async def resync_icd(self, node_id: int) -> None:
         """
         await self.send_command(APICommand.RESYNC_ICD, require_schema=12, node_id=node_id)
 
+    async def get_network_topology(self, refresh: bool = False) -> NetworkTopology:
```
Apollon77, line 312: Since HA core will consume this: worth stating explicitly in the docstring that `refresh=True` must not be polled — it triggers a per-node attribute read fan-out on the server, and (as of #832 today) concurrent callers are not coalesced, so N pollers cost N× the Thread/Wi-Fi traffic. The push event is the intended steady-state source; `refresh=True` is a user-initiated action.

Apollon77 (commented):
Reviewed alongside #832. This side is clean — the open-string-set choice for `kind`/`role`/`strength`, the unknown-key forward-compat test, and the per-command `require_schema=13` gate (baseline staying at 11) are all the right calls, and the schema-12 rejection test actually proves the gate rather than assuming it.

Two non-blocking notes:

1. Merge order: `require_schema=13` is hardcoded while `main` still reports `SCHEMA_VERSION = 12`, so this has to land after #832 (which bumps it and updates `python_client/tests/test_integration.py`). No file overlap between the two, so no conflict either way.
2. `refresh=True` is expensive server-side, and #832 currently has no in-flight coalescing (each caller fans out reads to every online node). Callers polling it would multiply that. Docstring note below.

Copilot's `strength`-in-docstring remark is already addressed in the current head.

python_client/matter_server/client/client.py
Apollon77: Added in 4272e5ee: docstring now states the refresh fan-out cost and that the push event is the source for keeping a graph current. Leaving open for review.

Issue #927, opened 2026-07-24
Invoke command responses use matter.js's own field-name casing instead of the python-matter-server/CHIP-clusters convention


lboue:
## Matter Server version
`origin/main` @ `dc09ef963649343bc68c5a69fff73120094d9932` ("v1.3.2-alpha.0-20260724-6944771 (#926)")

Also confirmed present in a local dev checkout pinning a slightly older `@matter/main` (`0.17.7-alpha.0-20260720-00d9d6fbf` vs. main's `0.17.7-alpha.0-20260723-8a6b27aed`) — same behavior on both.

## How do you run the Matter Server?
Standalone npm / node (`packages/ws-controller`)

## Node.js version
v24.18.0

## Operating system
Ubuntu 24.04.3 LTS (x86_64)

## Description

**What I did:** First hit this live: Home Assistant's Matter camera integration calls `CameraAvStreamManagement.VideoStreamAllocate` / `AudioStreamAllocate` (standard generic invokes) before requesting a WebRTC offer, then reads the allocated stream ID from the command response using the field name from the cluster's ClusterObjectFieldDescriptor (`videoStreamID` / `audioStreamID`, matching the CHIP SDK / python-matter-server convention) — this raised a `KeyError`. To get an isolated, reproducible repro without needing HA or a real device in the loop, I exercised the exact conversion function directly with debug logging (repro script below): it calls the same `ClusterMap` lookup and `convertMatterToWebSocketNameBased()` that `WebSocketControllerHandler#handleDeviceCommand` and `#handleSendWebRtcProviderCommand` use, feeding in a synthetic matter.js-shaped response for `VideoStreamAllocate`, `AudioStreamAllocate`, and `ProvideOffer`.

**What I expected:** The invoke response to use the key `videoStreamID` (and `audioStreamID`, `webRtcSessionID`), matching python-matter-server's own generic-invoke response convention — the WS API this server is meant to be a drop-in replacement for.

**What happened:** The response uses `videoStreamId` / `audioStreamId` / `webRtcSessionId` (lowercase `d`) instead, so a client following the python-matter-server convention gets a `KeyError` (or silently reads `undefined`) for the expected key.

### Analysis

[...]

PR #928, opened 2026-07-24
fix(converters): also expose CHIP SDK/python-matter-server ID casing in name-based invoke responses

lboue:
## Type of change

## Description

`convertMatterToWebSocketNameBased()` (`packages/ws-controller/src/server/Converters.ts`) builds each name-based command-response key from `member.propertyName` — matter.js's own field name — even though its doc comment says it exists "to match Python Matter Server behavior". matter.js hardcodes some fields with a lowercase `Id` suffix (e.g. `videoStreamId`) where the CHIP SDK / python-matter-server convention capitalizes it (`videoStreamID`), confirmed against this repo's own generated `python_client/chip/clusters/cluster_defs/*.py` typings. Clients following that convention (e.g. Home Assistant's Matter camera integration, allocating a stream via `CameraAvStreamManagement.VideoStreamAllocate` before a WebRTC offer) get a `KeyError` reading the response.

Per discussion on #927: rather than deriving/renaming the key from matter.js's own internal datatype references (the referenced type name isn't guaranteed to match the actual python-matter-server label for every field — the generated Python client typings are the real contract, not matter.js's internal model), this PR takes the simplest backward-compatible route and **exposes both keys**: the existing matter.js-cased entry is kept as-is, and a second entry with the trailing `Id` replaced by `ID` is added alongside it whenever a name-based struct field ends in `Id`. For `WebRtcSessionId` specifically this correctly yields `webRtcSessionID` (matter.js doesn't recognize "rtc" as an acronym at all, so only the trailing `Id`→`ID` flip applies — matching the Python client's `webRtcSessionID` label exactly). Tag-based (numeric-key) conversion is untouched.

**This is intentionally scoped to the reported ID-suffix case, not a general fix.** `python_client/scripts/generate-python-clusters.ts` already has a much more complete, battle-tested mechanism for this class of problem (`ACRONYMS`, `toChipName()`, `FIELD_NAME_OVERRIDES`, `K_VALUE_OVERRIDES` — covering MAC/URL/RFID/NOC/IPv4-IPv6/etc., plus dozens of one-off CHIP SDK spelling quirks) used to generate the very python_client typings this fix targets. A follow-up that extracts/shares that logic so `convertMatterToWebSocketNameBased` covers every cluster and acronym consistently (not just trailing `Id`) would be more correct long-term — this PR is the pragmatic stop-gap discussed in the issue, to be revisited "once the WS protocol is redocumented and officially switches to matter.js's own camelCase."

## Backing evidence

- Related issue: #927
- [x] This fix/change is backed by a **complete log file** — attached in #927 (`matterjs-server-casing-bug-debug.log` + repro script), reproducing the casing mismatch directly against the production `ClusterMap` → `convertMatterToWebSocketNameBased` code path for `VideoStreamAllocate`, `AudioStreamAllocate`, and `ProvideOffer`.

## Checklist

- [x] Tests added or updated to cover the change (`packages/matter-server/test/converter/ConvertersTest.ts`: dual-key assertions for `VideoStreamAllocateResponse`, `AudioStreamAllocateResponse`, `ProvideOfferResponse`; a non-`Id`-suffixed regression guard; a tag-based-conversion-unaffected guard)
- [x] `npm test` passes — `ws-controller` (241/241) and `matter-server`'s `Converters` suite (104/104) pass. The full `matter-server` suite has 2 pre-existing failures in `Integration Test › Device Discovery` (mDNS-dependent); reproduced identically against the unmodified base commit (`dc09ef9`) in a separate worktree, confirming they're environmental (no multicast in this sandbox), not caused by this change.
- [x] `npm run format-verify` and `npm run lint` pass (scoped to the changed files: `oxfmt --check` and `oxlint --type-aware` both clean)
- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This page shows a detailed overview of the changes between versions without the 
 
 ## **WORK IN PROGRESS**
 
+- Fix: (lboue) Name-based command responses (e.g. `VideoStreamAllocate`, `ProvideOffer`) now also expose the CHIP SDK/python-matter-server "ID" casing (e.g. `videoStreamID`) alongside matter.js's own casing (`videoStreamId`), so clients following either convention can read the response
 - Fix: Fixes the DST determination
 
 ## 1.3.1 (2026-07-23)
--- a/packages/matter-server/test/converter/ConvertersTest.ts
+++ b/packages/matter-server/test/converter/ConvertersTest.ts
@@ -969,6 +969,91 @@ describe("Converters", () => {
             expect(result).to.be.null;
         });
 
+        it("should also expose the CHIP SDK/python-matter-server ID casing for VideoStreamAllocateResponse (issue #927)", () => {
+            // CameraAvStreamManagement (1361), VideoStreamAllocateResponse has a single field whose
+            // matter.js property name is "videoStreamId" (lowercase d) but whose CHIP SDK / python-
+            // matter-server wire name is "videoStreamID" (capital ID). Rather than guess/rename,
+            // both keys are exposed for backward compatibility until the WS protocol is
+            // redocumented and officially switches to matter.js's own camelCase.
+            const cameraCluster = ClusterMap[1361]!;
+            const videoStreamAllocateCmd = cameraCluster.commands["videostreamallocate"]!;
+
+            const matterValue = { videoStreamId: 7 };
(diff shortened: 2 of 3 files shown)


Apollon77:
closing in favor of https://github.com/matter-js/matterjs-server/pull/930

PR #930, opened 2026-07-24
fix: align invoke/event wire field casing with Python client (#927)

### LINKED ISSUE #927: Invoke command responses use matter.js's own field-name casing instead of the python-matter-server/CHIP-clusters convention

Apollon77:
Fixes #927.

## Problem

Name-based WebSocket output (invoke command responses + events) keyed struct fields off matter.js `propertyName`, which lowercases acronym suffixes (`videoStreamId`, `pakePasscodeVerifier`). The Python Matter Server contract — which Home Assistant follows — spells these with preserved acronyms (`videoStreamID`, `PAKEPasscodeVerifier`).

The contract is not external: this repo generates the Python client (`python_client/`) from the matter.js model, and that generator already emits the correct labels. The server just used a different function for the same job.

## Fix

- **Single source of truth:** extracted the acronym-preserving field-name transform into `@matter-server/ws-client` (`wire-naming.ts`, pure strings). The Python-client generator and the WebSocket converter now import the *same* function, so wire keys and generated Python labels cannot drift (a byte-identical regenerate gate enforces this).
- **Dual-emit (non-breaking):** the name-based converter now emits both the acronym-corrected wire key AND the legacy `propertyName` key (same value) when they differ. Contract-conformant clients read the corrected key; anything that adapted to the old casing keeps working. A future major can drop the legacy key.
- Tag-based output (attributes, numeric keys) is untouched.

## Scope

43 field renames across command responses + events (`...Id`→`...ID` plus interior acronyms `PIN`/`PAKE`/`UTC`/`DST`/`NOCSR`/`ICD`/`URI`/`CSR`/`ARL`/`IP`). Inbound is already `camelize()`-tolerant, so both forms are accepted on the way in.

## Verification

- ws-client unit tests + converter dual-emit tests (incl. nested-struct) green.
- Full suite 741/741; Python parity gate empty at every step.
- Independent adversarial review over the cumulative diff: no correctness defects — swept 1092 structs / 3236 members / 463 dual-emit fields → 0 key collisions; round-trip confirmed.
- **End-to-end against a real Matter device:** integration test invokes `Groups.AddGroup` on the commissioned test light and asserts the wire response carries both `groupID` and `groupId` (255/255).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This page shows a detailed overview of the changes between versions without the 
 ## **WORK IN PROGRESS**
 
 - Fix: Fixes the DST determination
-- Fix: Command responses and events now expose acronym field names in the Python Matter Server casing (e.g. `videoStreamID`, `groupID`, `PAKEPasscodeVerifier`), matching the generated Python client and Home Assistant; the previous lowercased-acronym keys are still emitted alongside for compatibility (#927)
+- Fix: Command responses and events now expose acronym field names in the Python Matter Server casing (e.g. `videoStreamID`, `groupID`, `PAKEPasscodeVerifier`), matching the generated Python client and Home Assistant; the previous lowercased-acronym keys are still emitted alongside for compatibility
 
 ## 1.3.1 (2026-07-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This page shows a detailed overview of the changes between versions without the 
 ## **WORK IN PROGRESS**
 
 - Fix: Fixes the DST determination
+- Fix: Command responses and events now expose acronym field names in the Python Matter Server casing (e.g. `videoStreamID`, `groupID`, `PAKEPasscodeVerifier`), matching the generated Python client and Home Assistant; the previous lowercased-acronym keys are still emitted alongside for compatibility (#927)
 
 ## 1.3.1 (2026-07-23)
 
--- a/packages/matter-server/test/IntegrationTest.ts
+++ b/packages/matter-server/test/IntegrationTest.ts
@@ -562,6 +562,18 @@ describe("Integration Test", function () {
             await client.deviceCommand(commissionedNodeId, 1, 6, "off", {});
             await waitForOnOffUpdate(client, commissionedNodeId, false);
         });
+
+        it("should dual-emit acronym-cased command response fields (issue #927)", async function () {
+            // Groups.AddGroup (cluster 4) response carries GroupID: over the wire the payload
+            // must expose the python-matter-server casing (groupID) alongside the legacy key.
+            const result = (await client.deviceCommand(commissionedNodeId, 1, 4, "addGroup", {
+                groupId: 1,
+                groupName: "int-test",
+            })) as Record<string, unknown>;
+
+            expect(result.groupID).to.equal(1);
+            expect(result.groupId).to.equal(1);
+        });
     });
 
     // =========================================================================
--- a/packages/matter-server/test/converter/ConvertersTest.ts
+++ b/packages/matter-server/test/converter/ConvertersTest.ts
@@ -1179,4 +1179,38 @@ describe("Converters", () => {
             expect(result).to.deep.equal([2, 3]);
         });
     });
+
+    describe("convertMatterToWebSocketNameBased - acronym casing dual-emit (issue #927)", () => {
+        it("emits both the corrected and legacy key for an acronym field", () => {
+            // Groups cluster (4), AddGroup response = AddGroupResponse { 0:Status, 1:GroupId }
+            const groupsCluster = ClusterMap[4]!;
+            const addGroupCmd = groupsCluster.commands["addgroup"]!;
+            const responseModel = addGroupCmd.responseModel;
+
+            const result = convertMatterToWebSocketNameBased(
+                { status: 0, groupId: 5 },
+                responseModel,
+                groupsCluster.model,
+            ) as Record<string, unknown>;
+
+            expect(result.groupID).to.equal(5);
+            expect(result.groupId).to.equal(5);
+            expect(result.status).to.equal(0);
+        });
+
+        it("does not duplicate a key when corrected name equals propertyName", () => {
+            const groupsCluster = ClusterMap[4]!;
+            const addGroupCmd = groupsCluster.commands["addgroup"]!;
+            const responseModel = addGroupCmd.responseModel;
+
+            const result = convertMatterToWebSocketNameBased(
+                { status: 0, groupId: 5 },
+                responseModel,
+                groupsCluster.model,
+            ) as Record<string, unknown>;
+
+            // "status" is acronym-free: exactly one key, no alias
+            expect(Object.keys(result).filter(k => k.toLowerCase() === "status")).to.have.length(1);
+        });
+    });
 });
--- a/packages/matter-server/test/converter/ConvertersTest.ts
+++ b/packages/matter-server/test/converter/ConvertersTest.ts
@@ -1212,5 +1212,22 @@ describe("Converters", () => {
             // "status" is acronym-free: exactly one key, no alias
             expect(Object.keys(result).filter(k => k.toLowerCase() === "status")).to.have.length(1);
         });
+
+        it("dual-emits an acronym field nested inside a struct", () => {
+            // GroupKeyManagement cluster (63), KeySetRead response = KeySetReadResponse { 0: GroupKeySet }
+            // GroupKeySetStruct nests GroupKeySetId, so the dual-emit must reach the nested level too.
+            const gkmCluster = ClusterMap[63]!;
+            const keySetReadCmd = gkmCluster.commands["keysetread"]!;
+            const responseModel = keySetReadCmd.responseModel;
+
+            const result = convertMatterToWebSocketNameBased(
+                { groupKeySet: { groupKeySetId: 5 } },
+                responseModel,
+                gkmCluster.model,
+            ) as Record<string, Record<string, unknown>>;
+
+            expect(result.groupKeySet.groupKeySetID).to.equal(5);
+            expect(result.groupKeySet.groupKeySetId).to.equal(5);
+        });
     });
 });
--- a/packages/ws-client/package.json
+++ b/packages/ws-client/package.json
@@ -37,7 +37,8 @@
     "scripts": {
         "clean": "nacho-build clean",
         "build": "nacho-build",
-        "build-clean": "nacho-build --clean"
+        "build-clean": "nacho-build --clean",
+        "test": "matter-test --force-exit"
     },
     "dependencies": {
         "@matter/thread-br-client": "0.17.7-alpha.0-20260723-8a6b27aed"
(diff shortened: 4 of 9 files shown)


Apollon77:
Thanks @copilot — validated both against the codebase:

**WireNamingTest import (valid — fixed in 6b34333b):** switched to `import … from "../src/wire-naming.js"` to match the sibling ws-client tests (`WsClientTest.ts`, `TopologyUtilsTest.ts`), which all import from `../src/*` rather than the package entrypoint. Tests green (79/79).

**Generator importing `@matter-server/ws-client` (won't change):** this doesn't introduce a new clean-checkout dependency. `generate-python-clusters.ts` already requires a built workspace — it imports `@matter-server/custom-clusters` (line 20) and `../../packages/custom-clusters/dist/esm/clusters/index.js` (line 28). `npm run python:generate` is a post-build step by design, and CI runs `npm run build` before it (`test-python-client.yml`). Importing the helper via a relative `src` path instead would bypass the package's public boundary without removing the build requirement (custom-clusters' dist is still needed), so it's a net loss. Keeping the clean package import.

PR #935, opened 2026-07-26
network-details.ts: harmonize casing of labels

schildbach:
This change is entirely cosmetic.

Extracts a common style from the existing labels and corrects the outliers.

- First word starts uppercase, consecutive words lowercase
- Acronyms use all-caps, except where there is an established camelcase


Apollon77:
thanks

PR #939, opened 2026-07-26
Add dashboard upload for local OTA firmware files

lboue:
## Type of change

## Description

Adds a way to upload a local `.ota` Matter firmware image from the dashboard, without needing CLI/filesystem access to the server.

- New `initiate_ota_upload` WebSocket command reserves a random, single-use upload id (and one of a limited number of in-flight slots) before any upload starts, valid for 60s.
- New `POST /ota-upload/<upload_id>` HTTP endpoint streams the raw binary body straight to a temp file (checked against the configured size limit as it arrives), then imports it into the server's local OTA image store — the same store populated by `--otaProviderDir` directory scanning. Returns the parsed vendor ID / product ID / software version as JSON. Kept as a plain HTTP endpoint (not a WebSocket command) so large firmware images don't pay the ~33% base64 overhead or hit the WebSocket library's default 100 MiB message cap, and so the body never has to be buffered whole in memory.
- New "Upload Firmware" button on a node's detail page in the dashboard, calling `initiate_ota_upload` then POSTing the selected file directly. After a successful upload, if the parsed vendor/product ID matches the current node, it chains into the existing check/update-node flow; otherwise it shows an informational dialog, since an `.ota` image isn't node-scoped — it's identified by vendor/product ID in its header and becomes discoverable by any matching node via the existing `check_node_update`.
- New CLI flags `--ota-upload-max-in-flight` (default 5) and `--ota-upload-max-size-mb` (default 64) to tune the concurrency/size limits; the existing `--disable-ota` also gates the new upload command/endpoint.
- New OHF-extension error code `OtaUploadError` (101) for corrupt images, an unknown/expired/already-used upload id, disabled OTA support, or store failures.
- Startup orphan cleanup only removes files matching the upload naming scheme (`<32-hex>.ota`) it actually generates, so a misconfigured `tempDir` can't wipe unrelated data.

This addresses the design [@Apollon77 proposed in review](https://github.com/matter-js/matterjs-server/pull/939#pullrequestreview-4866321492): declaring the upload over the (already-authenticated) WebSocket session before the HTTP POST is accepted, streaming to disk instead of buffering in memory, and bounding concurrent uploads.

## Backing evidence

N/A — this is a feature addition, not a fix for an observed defect.

- Related issue: #

## Checklist

- [x] `npm test` passes (only pre-existing documented mDNS/commissioning integration flakes remain — see `CLAUDE.md`)
- [x] CHANGELOG updated


Apollon77:
Hm ... I am really unsure if i want to push some Megabyte of data through Websocket ... need to think about this

lboue:
No problem. Come to think of it, the idea is to make it easier for users to upload update files.

Apollon77:
I think it is better to use HTTP techniques to transfer 1-25MB Matter OTA images (and yes they can get bigger when you consider cameras and other device types). Doing that via websocket means that we need to have the full data in RAM because we can not really "stream it" and I think we should avoid this.

Apollon77 (changes requested):
see last comment

docs/websockets_api.md
```diff
@@ -774,6 +774,33 @@ Import nodes from Home Assistant diagnostic dumps for testing purposes. Test nod
...
 ```
 
+## HTTP Endpoints
+
+Some functionality is exposed over plain HTTP instead of the WebSocket command channel,
+served by the same listener/port as `/ws`.
+
+**POST /ota-upload** - Store a local `.ota` firmware file in the OTA image store
+
+The request body is the raw `.ota` file bytes (no base64/JSON envelope), with an optional
```
Apollon77: Thats not rigth ... we shoukd really use HTTP POST multipart streaming here ...

packages/matter-server/src/server/OtaUploadHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+                if (aborted) return;
+                size += chunk.length;
+                if (size > MAX_OTA_UPLOAD_SIZE) {
+                    aborted = true;
+                    res.writeHead(413, { "Content-Type": "application/json" });
+                    res.end(JSON.stringify({ error: "File too large" }));
+                    req.destroy();
+                    return;
+                }
+                chunks.push(chunk);
```
Apollon77: The whole idea of using HTTP style uploads is that we can also process this streaming and write info the file. Then we use the file on disk to parse the data streaming again. Not having all data in memory

packages/matter-server/src/server/OtaUploadHandler.ts
```diff
@@ -0,0 +1,96 @@
...
+            }
+            if (req.method !== "POST") {
+                res.writeHead(405, { Allow: "POST" });
+                res.end();
+                return;
+            }
+
+            let fileName: string | undefined;
+            try {
+                fileName = new URL(req.url, "http://localhost").searchParams.get("file_name") ?? undefined;
```
Apollon77: no need for any filename ... we should simply upload as a random filename because we process and delete anyway directly afterwards and only keep it if it is a valid ota file

packages/ws-controller/src/controller/ControllerCommandHandler.ts
```diff
@@ -1790,6 +1791,43 @@ export class ControllerCommandHandler {
...
+     */
+    async uploadOtaFile(data: Uint8Array, fileName?: string): Promise<MatterSoftwareVersion> {
+        if (!this.#otaEnabled) {
+            throw ServerError.otaUploadError("OTA is disabled");
+        }
+        try {
+            const otaService = await this.#controller.node.act(agent => agent.get(DclBehavior).otaUpdateService);
+            await otaService.construction;
+
+            const otaUrl = `local://${encodeURIComponent(fileName ?? "upload")}`;
```
Apollon77: is "local://" really correct here? Should be "file:"?

Apollon77 (changes requested):
I added some comments ... in the end I wouldalso propose for the sake of security to change the process a bit. WHat if we define one websocket command (schema 13) whichis "initialize_ota_upload" (kind of, not final name). a call is returning an "upload identifier" (could also be the filename in second step) and registers this at the upload handler to start a transfer within timeframe of lets say 60s. then the POST for the data needs to have this id instead the "filename" and only works when it is registered and found. So we ensure that any upload got declared via the ws api which might be passwort or otherwise "protected". we also need a CLI flag to disable the upload capability because formally it widens the security attack vector

docs/websockets_api.md
```diff
@@ -774,6 +774,33 @@ Import nodes from Home Assistant diagnostic dumps for testing purposes. Test nod
...
+`file_name` query parameter used to namespace the local image. The image is stored by vendor
+ID / product ID / software version parsed from its header, not tied to any particular node —
+`check_node_update` will surface it for any node whose vendor/product matches.
```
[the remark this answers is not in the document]
lboue: Superseded by the redesign in commit 9d996be and earlier: the shipped interface is the `initiate_ota_upload` WebSocket command + `POST /ota-upload/<upload_id>`, not the old `upload_ota_file` base64 command this comment refers to. The PR description and `docs/websockets_api.md` already describe the current shape. Resolving.

packages/ws-controller/src/controller/OtaUploadRegistry.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        for (const entry of entries) {
+            try {
+                await rm(join(this.#tempDir, entry), { force: true, recursive: true });
+                logger.info(`Removed orphaned OTA upload ${entry}`);
+            } catch (error) {
+                logger.warn(`Failed to remove orphaned OTA upload ${entry}:`, error);
+            }
+        }
```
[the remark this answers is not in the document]
lboue: Fixed in 9d996be: `cleanupOrphans()` now only removes entries matching the `<32-hex>.ota` naming scheme `reserve()` actually generates, and dropped `recursive: true` since matched entries are always plain files. Added a regression test asserting unrelated files/dirs in `tempDir` are left alone.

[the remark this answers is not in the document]
lboue: Same fix, see 9d996be — duplicate of the earlier comment on this line.

lboue:
Ready for the review

PR #941, opened 2026-07-27
fix: re-register observers on sleepy device reconnection

### LINKED ISSUE #940: Events stop flowing from sleepy devices (ICDs) after subscription re-establishment
## Summary

Events from sleepy end devices (ICDs) stop being forwarded to Home Assistant after the device wakes up from a long idle period. Restarting matter-server restores event flow temporarily, but it breaks again after the next idle/reconnect cycle.

## Environment

- matterjs-server: stable (ghcr.io/matter-js/matterjs-server:stable)
- Device: IKEA BILRESA scroll wheel (Thread sleepy end device with 15-minute idle interval)
- Home Assistant Matter integration

## Reproduction

1. Commission a sleepy Thread device (e.g., IKEA BILRESA, RODRET, SOMRIG)
2. Interact with the device - events flow to HA correctly
3. Leave the device idle for longer than its idle interval (15 minutes for BILRESA)
4. Interact with the device again after it reconnects
5. Events no longer appear in HA, even though matter-server logs show "Received event"

## Root Cause Analysis

When a sleepy device's subscription times out and matter.js re-establishes it, the underlying event emitters on the `PairedNode.events` object may be replaced. The observers registered at startup by `ControllerCommandHandler#registerNode` become orphaned - they're still stored in `#nodeObservers` but no longer connected to the live event emitters.

This causes `Observable.emit()` to silently return early (no observers to call), and events never reach `WebSocketControllerHandler`.

## Evidence

- `ClientEventEmitter` logs show events being received by matter.js
- `WebSocketControllerHandler` never receives the forwarded events
- Restarting matter-server works because it re-registers all observers
- Fast reconnects (within the 3-minute grace period) don't exhibit this issue

## Fix

The fix is to re-register observers when a node transitions to `Connected` state after being in `Reconnecting` state:

1. Extract observer registration into a reusable `#setupNodeObservers()` method
2. Call it on every non-fast reconnection in `#handleNodeStateChange()`

PR with fix: https://github.com/MarkAtwood/matterjs-server/tree/fix/sleepy-device-observer-reconnect

## Workaround

Restart matter-server after sleepy devices become unresponsive:
```bash
docker restart matter-server
```

MarkAtwood:
## Summary

Fixes #940 - Events from sleepy end devices (ICDs) stop flowing to Home Assistant after reconnection.

## Problem

When a sleepy device reconnects (e.g., IKEA BILRESA with 15-minute idle interval), matter.js may replace the underlying event emitters on `PairedNode.events`, orphaning the observers registered at startup. This causes events to be received by matter.js but never forwarded to Home Assistant.

## Solution

1. Extract observer registration into a reusable `#setupNodeObservers()` method that safely closes any existing observers before creating new ones
2. Call this method on **every** reconnection in `#handleNodeStateChange()` to re-wire observers to the current event emitters

Initially the fix only triggered on "slow" reconnects (after the 3-minute grace period), but testing revealed that sleepy devices can orphan observers even on fast reconnects. The cost of re-registration is minimal compared to silently dropped events.

## Testing

- All 241 existing tests pass
- Manual testing with IKEA BILRESA scroll wheel shows events continue to flow after multiple reconnection cycles, including fast reconnects

## Changes

- `ControllerCommandHandler.ts`: Added `#setupNodeObservers()` method and unconditional re-registration on reconnection

### DIFF
--- a/packages/ws-controller/src/controller/ControllerCommandHandler.ts
+++ b/packages/ws-controller/src/controller/ControllerCommandHandler.ts
@@ -541,11 +541,20 @@ export class ControllerCommandHandler {
         return this.#controller.close();
     }
 
-    async #registerNode(nodeId: NodeId) {
-        const node = await this.#controller.getNode(nodeId);
+    /**
+     * Set up event observers for a node. Creates a new ObserverGroup and registers all event
+     * handlers. Safe to call multiple times — closes any existing observers before creating new ones.
+     *
+     * Sleepy end devices (ICDs) can lose their subscription during long idle periods. When matter.js
+     * re-establishes the subscription, the underlying event emitters may be replaced, orphaning old
+     * observers. This method re-wires observers to the current event emitters.
+     */
+    #setupNodeObservers(node: PairedNode): void {
+        const nodeId = node.nodeId;
         const attributeCache = this.#nodes.attributeCache;
 
-        // Per-node ObserverGroup so all subscriptions are cleaned up on decommission
+        this.#nodeObservers.get(nodeId)?.close();
+
         const nodeObservers = new ObserverGroup();
         this.#nodeObservers.set(nodeId, nodeObservers);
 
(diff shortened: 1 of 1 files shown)


Apollon77:
I close here, because IF this is really a observer issue (which I really do not think!) then we need to fix it in matter,js ... so back to "add logs to the issue" instead of fixes without any proof

PR #952, opened 2026-07-30
feat(dashboard): show client-mode clusters on endpoint view

lboue:
## Type of change

## Description

The endpoint cluster list in the dashboard was derived purely from cached attribute keys
(`node.attributes`), so client-mode clusters (an endpoint binds to a cluster hosted on another
device rather than hosting it itself) never appeared, even though the Descriptor's `ClientList`
attribute records them. Client clusters have no local attribute storage, so they were silently
invisible with no explicit filter excluding them — it was an emergent effect of the data
collection method.

**Update after review:** the first version of this PR merged `ClientList` cluster ids directly
into the Clusters list with a `CLIENT` badge, but as [@Apollon77 pointed out](https://github.com/matter-js/matterjs-server/pull/952#issuecomment-5143329334),
those entries weren't clickable and did nothing when tapped. Reworked per that feedback:

- Client-mode clusters are no longer added to the Clusters list.
- Instead, `matter-endpoint-view.ts` shows a "Client Clusters" chip panel near the top of the
  page — but **only when the endpoint also has a Binding cluster**, since without one there's
  nothing to do with a client cluster anyway.
- Clusters already covered by an existing binding entry (`util/binding.ts`'s new
  `boundClientClusterIds`, cross-referencing the Binding cluster's `binding` attribute against
  the Descriptor's `ClientList`) are visually marked "Bound" vs. the rest.
- The cluster detail view (`matter-cluster-view.ts`) still shows a short notice explaining why a
  client-mode cluster's attribute list is empty, for anyone navigating there directly.

## Backing evidence

Not applicable — this is a feature addition, not a fix for a defect.

- Related issue: #

## Checklist

- [ ] Tests added or updated to cover the change
- [x] `npm test` passes

---

Manually verified against a running server: endpoint 38 (Thermostat Controller) has
`ServerList = [29, 57, 30]` and `ClientList = [513]` (Thermostat). After this change, the
Clusters list no longer shows the Thermostat client cluster; instead, because the endpoint has
the Binding cluster (30), a "Client Clusters" panel appears showing the Thermostat cluster as a
chip, marked "Bound" once a binding entry targets it.

No new automated tests were added for this UI change; verification was manual (see above) plus
the existing dashboard test suite (153/153 passing).


Apollon77:
@lboue Hm, honestly I do not really see the benefit of having the Client clusters also in the Cluster list, You can not click on then and basically nothing happens with them.
What about listing them. in the section on top of the page where we also list Tags and features? Kind of a list as "chips" list wit "Client clusters" ... I would also maybe only list them when the EP also has a bindings cluster because if not you can not do anything with them anyway. And if we would like to be very convenient we can read the bindings and mark the Cluster clients that are "Bound" somehow differently then the others.

Apollon77 (dismissed):
see last comment

lboue:
@Apollon77 Good points, thanks. Reworked in fd7ce21:

- Client-mode clusters are no longer merged into the Clusters list (the unclickable `CLIENT` badge entries are gone).
- They now show as chips in a new "Client Clusters" panel near the top of the endpoint view, and only when the endpoint also has a Binding cluster — if there's no Binding cluster, the panel doesn't render at all.
- Clusters already covered by an existing binding entry are marked "Bound" (cross-referencing the Binding cluster's `binding` attribute against the Descriptor's `ClientList`, in `boundClientClusterIds()`).

Verified manually against the same Thermostat Controller endpoint (`ServerList = [29, 57, 30]`, `ClientList = [513]`): the Thermostat cluster no longer appears in the Clusters list, and shows up as a chip in the Client Clusters panel, marked "Bound" once a binding targets it.

Apollon77:
Pushed two maintainer edits: `b845bdcc` (the client-cluster panel itself) and `e62fd741` (fabric-filtering the reads).

**Dropped the cluster-detail notice.** Nothing links to a client-mode cluster page — the Clusters list comes from cached attribute keys and the chips aren't links — so it was only reachable via a hand-typed URL, and in the one case where the page *is* reachable (a cluster in both ServerList and ClientList) the attribute check suppressed it anyway.

**The panel itself:** cluster ids on the chips, Bound/Not bound as text rather than colour plus a tooltip, the duplicated panel/chip CSS extracted into `shared-styles`, and the missing light-theme `--md-sys-color-on-secondary-container` added (that also fixes the existing cluster-view chips). `guard` deps in the endpoint view now include `node_id`, `endpoint` and the attributes object identity — your `Object.keys(...).length` fix was right, the old `.length` on a Record was always `undefined`, but the element is reused across route changes so the count alone let the previous endpoint's list persist.

**Then the interesting part.** Marking a cluster "Bound" turned out to be reading contaminated data. matter.js only integrates a one-shot read into the subscribed datasource when the read's fabric filtering matches the subscription's (`ClientStructure#mutateAttribute`) — an unfiltered read returns values the server deliberately never caches. But the dashboard was merging those responses into its own `node.attributes`: the binding editor, the ACL panel and the dev-mode attribute refresh all did. On a node commissioned into a second ecosystem that left foreign entries in a cache nothing ever refreshes.

Two consequences beyond the chips:

- The binding editor's `render()` uses unfiltered `readBindings()` while `_delete(i)` passes that positional index to `deleteBindingAtIndex`, which *does* filter by fabric. Phantom foreign rows shifted the index — deleting row N could remove a different binding.
- `freshOurAcl` / `freshOurBindings` rewrite these list attributes whole. A read reports a per-path failure by omitting the path (`handleReadAttributes` warns and drops it; the WS handler only throws when *every* path fails), so an absent `0/31/0` became `attributeArray(undefined)` → `[]` → a write of an empty ACL, locking every controller out of the node. They now abort instead, with tests.

Every dashboard read that feeds the cache is fabric-filtered now. The one exception is ICD's `_otherClientCount`, which must count other fabrics — it stays unfiltered but no longer merges that value into the cache.

**Two pre-existing bugs found on the way, not fixed here** — will file them separately:

1. **The ICD "other ecosystems are still registered" gate can never fire.** CHIP marks every non-`FabricIndex` field of `MonitoringRegistrationStruct` fabric-sensitive, so on an unfiltered read a foreign entry arrives as `{"254": N}` and nothing else. `decodeRegisteredClient` requires fields `1` and `2`, drops it, and `otherFabricClientCount` returns 0 every time — the warning dialog before disabling Battery Saver Mode is dead code, and `IcdUtilTest.ts` labels that exact wire shape "malformed" and asserts it is skipped.
2. **ACL/binding write results are ignored** — the returned `AttributeWriteResult[]` is discarded, so a ConstraintError write reports success in the UI.

One deliberate behaviour change worth flagging: the dev-mode per-attribute refresh is now fabric-filtered too, so a developer can no longer use it to see other fabrics' entries in `Fabrics`, `NOCs`, `GroupKeyMap` or `ACL`. Cache consistency won here; an explicit "unfiltered read" affordance that displays without merging would be the better long-term answer.

Full gate green: format, lint, build, `npm test` (integration included). Dashboard suite 164/164 — nine new tests, each mutation-verified.

Re the Copilot review on this PR: adopted the `BINDING_KEY_RE` / `BINDING_CLUSTER_ID` drift point. The other three don't hold — the guard already depends on the attributes object identity, not only the key count (nothing deletes attribute keys in place; removal arrives via `node_updated` with a fresh object), and returning an empty set when `CurrentFabricIndex` is uncached would be the more wrong answer now that the cache is fabric-clean by construction: entries present without a known index are ours, so "Not bound" would contradict reality. `entriesForFabric(entries, undefined)` returns all entries for the same reason, and the write path stays authoritative either way.

Apollon77:
@lboue please have a final check. for me looks good. when for you too we can merge

lboue:
That's fine with me, too
image

PR #954, opened 2026-07-31
fix(python_client): classify a node from Thread diagnostics on any endpoint

LorbusChris:
## Type of change

## Description

`MatterClient.node_diagnostics()` derives the node type only when the node is reached **over Thread**, and only from the `ThreadNetworkDiagnostics` cluster on **endpoint 0**:

```python
if network_type == NetworkType.THREAD:
    thread_cluster = node.get_cluster(0, Clusters.ThreadNetworkDiagnostics)
```

A Matter Network Infrastructure Manager (NIM) can satisfy neither condition, on the same device:

- it is commonly attached over **Ethernet or Wi-Fi**, so `network_type` is not `THREAD` and the whole block is skipped; and
- the Network Infrastructure Manager (`0x0090`) and Thread Border Router (`0x0091`) device types both **require** Thread Network Diagnostics on their own endpoint, and both are `Simple`, endpoint-scoped device types, so that endpoint is never the root one. On the Root Node device type (`0x0016`) the same cluster is optional and present only when the node is itself attached over Thread.

So `node_type` stays `UNKNOWN`, and Home Assistant shows no device type for a node that is in fact a Thread router or leader.

This takes the cluster from whichever endpoint provides it — lowest endpoint id first, so a device that reports on the root endpoint is unaffected — and derives the node type from its routing role regardless of how the node itself is attached. Two details preserved deliberately:

- the Wi-Fi `END_DEVICE` fallback now applies only when Thread said nothing, so a Wi-Fi-attached border router is described by its mesh role rather than flattened to an end device;
- `network_name` stays reserved for nodes actually attached over Thread — an Ethernet-attached border router should not report the Thread network as its own.

## Backing evidence

Tested and confirmed on an OpenWrt/Turris border router running the Matter `network-manager-app`: reached over Ethernet, with `ThreadNetworkDiagnostics` on the application endpoint beside `ThreadBorderRouterManagement`.

In the Home Assistant device page, under **Matter node diagnostics**:

| | Device type |
|---|---|
| Before | `Unknown` — the router is a Thread **leader**, but nothing in the UI says so |
| After | `Routing end device` |

Nothing else in that panel changes: the node still reports `Network type: Ethernet`, and the Thread network name stays empty for it, since the router is not itself attached over Thread.

The new test builds a node with an operational **Ethernet** interface on endpoint 0 and `RoutingRole = kLeader` on endpoint 1, and asserts `ROUTING_END_DEVICE`. It **fails on `main`** and passes with this change. The two tests covering the existing paths (Thread node classified from the root endpoint, Wi-Fi node without Thread) stay green, so the behaviour change is limited to the case above.

- Related issue: none yet, happy to open one if you prefer.

Left unchecked on purpose: I haven't attached a log. What backs this is my own deployment plus a test that fails without the fix. Happy to attach a diagnostics dump if you want the raw node data.

## Checklist

- [x] `npm test` passes

Also run for the Python side: `npm run python:test` (55 passed, against 52 on `main` — the 3 new tests) and `npm run python:lint`, plus `ruff format --check`, `mypy`, `codespell` and `pylint` (10.00/10). The 4 `test_client_integration.py` errors in my environment occur identically on `main`: the Node mock server does not start here.

---

Written with AI assistance (Claude). I have reviewed every line, verified the behaviour on my own deployment, and will answer review feedback myself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ This page shows a detailed overview of the changes between versions without the 
 	## **WORK IN PROGRESS**
 -->
 
+## **WORK IN PROGRESS**
+
+- Fix: (LorbusChris) Python client: `node_diagnostics` classifies a node from the Thread diagnostics of whichever endpoint provides them, instead of only the root endpoint of a node attached over Thread, so a border router reached over Ethernet or Wi-Fi is reported by its routing role rather than as an unknown node type
+
 ## 1.3.3 (2026-07-28)
 
 - Enhancement: (pkese) Dashboard network graphs space nodes by signal quality (Thread LQI, Wi-Fi RSSI) instead of using one fixed edge length
--- a/python_client/matter_server/client/client.py
+++ b/python_client/matter_server/client/client.py
@@ -463,29 +463,42 @@ class MatterClient:
         # get thread/wifi specific info
         node_type = NodeType.UNKNOWN
         network_name = None
-        if network_type == NetworkType.THREAD:
-            thread_cluster: Clusters.ThreadNetworkDiagnostics = node.get_cluster(0, Clusters.ThreadNetworkDiagnostics)
-            if thread_cluster:
+        # A node does not have to be attached over Thread to be part of a
+        # Thread network, and it does not have to serve the diagnostics from
+        # the root endpoint: a border router is commonly reached over Ethernet
+        # or Wi-Fi and reports Thread from the endpoint that carries its
+        # border router device type. Take the cluster from wherever it is, so
+        # such a node is still described by the role it plays in the mesh.
+        thread_cluster: Clusters.ThreadNetworkDiagnostics | None = next(
+            (
+                cluster
+                for endpoint_id in sorted(node.endpoints)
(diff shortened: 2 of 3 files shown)


Apollon77:
Hm ... this is difficult. The matter server has internally a very complex logic to determine the "devices network type" and this is not simply the existence of clusters but also needs to base on data of these. Especially when also looking at other endpoints and so potentially considering a device. with multiple network interfaces. And NIMs are then an even more special category.

The Diagnostics here also try to determine the "main device network interface" and for a NIM that has Thread "additionally" i do not know what exactly we should show where.

So with all this in fact we are exactly at the point where the PR templated asks for details, logs, and in your case, a Diagnostic JSON or other device structure details to understand what exactly you miss and want to achieve.

In fact because of the fact that the matter server already does all this determination internally I also could think about removing this special logic from the Python client and feed this via Node details directly from the matter server.

So for me this sounds like a topic we should discuss the goals and details before having a PR. I close this one. Please open an issue with all details and then we can discuss the best solution.

Issue #957, opened 2026-08-01
write_attribute: struct/list-of-struct writes fail with Status 141 (INVALID_DATA_TYPE) due to name-keyed vs tag-keyed JSON mismatch


lboue:
### Matter Server version

matter-server/1.3.3 (matter.js/0.17.7) — from the live `server_info` handshake (`sdk_version` field).

### How do you run the Matter Server?

Standalone (remote host on the LAN, reached over `ws://<host>:5580/ws`). Exact process manager not verified from this report (no shell access to that host from where this was investigated); local source checkout used for code inspection reports `0.0.0-git` in `packages/matter-server/package.json`, so this may be a from-source/dev deployment.

### Node.js version

Not captured from the running server. Local source checkout (used only to read/trace the code, not to run the server) is on Node v24.18.0.

### Operating system

Unknown — remote host, not directly inspected for this report.

### Description

**What I did:** Using a Python client (`matter-python-client` 1.3.0, from this project's `python_client` package) against a live Matter Server, called `write_attribute()` to write the Thermostat cluster's `Presets` attribute (endpoint 1, cluster 0x0201/513, attribute 0x50/80) as part of implementing the Matter `PRES`/`MSCH` (Presets/Schedule Configuration) atomic-write client flow. The target device is the connectedhomeip reference `thermostat-app` example (vendor 0xFFF1 / product 0x8001), which recently gained full `Presets`/`Schedules` support server-side in [project-chip/connectedhomeip#73342](https://github.com/project-chip/connectedhomeip/pull/73342).

**What I expected:** The write to succeed (status 0), or at least be rejected for a content-related reason (e.g. `ConstraintError` if violating a spec rule).

**What happened:** The write is rejected with `Status: 141` (`INVALID_DATA_TYPE`) **even when writing back the exact, unmodified list already reported by the device** — i.e. this is not about the content of the new value, it reproduces with a byte-for-byte round-trip of existing data. A companion write to a *scalar* attribute (`OccupiedHeatingSetpoint`, same cluster/endpoint) succeeds normally (`Status: 0`) through the same connection, isolating the problem to attributes whose value is a **list of structs**.

[...]

PR #958, opened 2026-08-01
fix(python-client): serialize write_attribute struct values by TLV tag, not field name

lboue:
## Type of change

## Description

`write_attribute()` in the Python client passed dataclass values (e.g. `list[PresetStruct]` for the Thermostat `Presets`/`Schedules` attributes) straight through to the outgoing WebSocket message, relying on `orjson`'s native dataclass serialization. `orjson` keys a dataclass by its Python **field name** (e.g. `"presetHandle"`), but the server's `WRITE_ATTRIBUTE` handler (`convertWebSocketTagBasedToMatter` in `ws-controller`) expects struct members keyed by their numeric **TLV tag** (`"0"`), matching the format already used for attribute reports.

Because of that mismatch, every struct field falls into an "unknown key, keep as-is" fallback that skips type conversion — most importantly, `octstr` handle fields (`presetHandle`/`scheduleHandle`) stay as raw base64 strings instead of being decoded to binary. matter.js then rejects the resulting malformed struct with `Status 141` (`INVALID_DATA_TYPE`) when trying to TLV-encode it. This reproduces even for a byte-for-byte unchanged round-trip write (confirmed against a live device), and is unrelated to the `AtomicRequest` Begin/Commit sequence, which both report success independently.

**Fix:** adds `dataclass_to_tag_dict()` (`common/helpers/util.py`), which recursively converts a dataclass instance — including nested lists, e.g. `Schedule.transitions` — into a dict keyed by each field's `ClusterObjectFieldDescriptor.Tag`, and uses it in `write_attribute()`. `dataclass_to_dict()` (field-name-keyed, used by `send_device_command` for `DEVICE_COMMAND` payloads) is untouched, since the server's command-invoke path (`convertCommandDataToMatter`) intentionally expects name-keyed data — the two message types use different wire conventions on the server side.

## Backing evidence

- Related issue: #957
- [x] This fix/change is backed by a **complete log file** — attached here or in the linked issue (not just a few quoted lines).

Issue #957 contains the exact raw WebSocket request/response JSON captured directly from the wire for the failing write (both the malformed name-keyed request and the server's `Status: 141` response), plus the precise server-side source lines (`Converters.ts`, `ControllerCommandHandler.ts`) that explain why that shape is rejected. That is the complete, reproducible evidence this fix is based on.

## Checklist

- [x] Tests added or updated to cover the change (`python_client/tests/test_util.py`)
- [ ] `npm test` passes — not applicable, this change touches only `python_client/` (pure Python), no TypeScript packages
- [ ] `npm run format-verify` and `npm run lint` pass — not applicable for the same reason; ran the Python equivalents instead: `pytest python_client/tests/` (130 passed) and `ruff check`/`ruff format --check` on all changed files (clean, aside from a pre-existing repo-wide `CPY001` copyright-notice warning present on every file in the project, unrelated to this change)
- [x] CHANGELOG updated


Apollon77:
Pushed a follow-up commit on top of this PR — thanks @lboue, both sides of the fix were sound. What it adds:

**Null semantics** — the main gap: unset optional fields traveled as explicit `null` (orjson emits `null` for every unset dataclass field; the new tag-keyed helper did too). matter.js rejects `null` on optional non-nullable members, so partial-struct writes (e.g. a heat-only schedule transition) still failed. Now:
- Server: `WRITE_ATTRIBUTE` struct conversion treats `null` for an optional non-nullable member as omitted — same guard the command-invoke path already had. This is what actually unblocks pre-1.3.0 name-keyed clients.
- Python client: `dataclass_to_tag_dict()` omits `None` fields (matching CHIP's TLV encoder); `NullValue` still travels as `null`.

**Key parsing hardening** — only purely-numeric keys are treated as TLV tags (`"5x"` no longer resolves to tag 5), and the name fallback also accepts legacy matter.js property names (e.g. `iPv4Addresses`) which the server dual-emits since #927.

**Tests** — converter unit tests for the null/key rules (incl. a wire-name≠propertyName case via `GeneralDiagnostics.NetworkInterface`), a client test pinning the tag-keyed `write_attribute` payload, and end-to-end struct-list write round-trips in both integration suites (tag-keyed and name-keyed) via a new `UserLabel` cluster on the test light device.

All gates green: `npm run format` / `lint` / `build` / full `npm test` (incl. both integration suites), `pytest` 132 passed, `ruff` clean.

Apollon77 (approved):

Issue #962, opened 2026-08-02
Feature: Dashboard cluster view — Schedule Configuration panel for Thermostat (MSCH)


lboue:
## Summary

Add a dedicated panel to the Thermostat cluster view showing the `MatterScheduleConfiguration` feature's schedules (`Schedules` attribute) and active schedule (`ActiveScheduleHandle`).
This will allow us to run tests before the feature is supported in Home Assistant.

## Motivation

Thermostats supporting the MSCH feature currently expose `Schedules` / `ActiveScheduleHandle` as raw attribute values in the dashboard, which is hard to read at a glance. A visual weekly grid makes the setpoint pattern immediately understandable without decoding the `ScheduleStruct` / `ScheduleTransitionStruct` payloads manually.

## Proposed UI

- Only rendered when the cluster's FeatureMap includes `MatterScheduleConfiguration` (same conditional pattern as the Active Features panel).
- Chips row listing each schedule (name + handle), with an "Active" badge on the one matching `ActiveScheduleHandle`.
- A 7-day × 30-min grid, color-coded by heat setpoint, for the selected schedule.
- A transition list below (time / label / setpoint) for the selected schedule.

Mockup attached (screenshot) — read-only for a first iteration; a follow-up could add a `SetActiveScheduleRequest` action button and, longer-term, an editable graph view (nodes draggable on a time/temperature curve, similar in spirit to community tools like Home Assistant's climate-scheduler card) if we want the dashboard to support editing schedules rather than just visualizing them.

<img width="620" height="469" alt="Image" src="https://github.com/user-attachments/assets/ff30b028-a908-41d0-bdfc-867ced87930a" />

### Confirmation

- [ ] If this request is about a current limitation or wrong behavior, I attached a log file showing it.
- [x] I have reviewed and understand everything in this request and can explain it in my own words ([AI policy](https://github.com/matter-js/matterjs-server/blob/main/AI_POLICY.md)).
- [x] I searched existing issues and this is not a duplicate.

Apollon77:
included in 1.4.0

PR #963, opened 2026-08-02
feat(dashboard): add Schedule panel for Thermostat (MSCH)

lboue:
## Summary
- Adds a Schedule Configuration panel to the Thermostat cluster view, gated on the FeatureMap's MSCH bit (same conditional pattern as the existing Active Features panel), closing #962.
- Visualizes the `Schedules` / `ActiveScheduleHandle` attributes as schedule chips (with an Active badge), a 7-day × 24h color-coded grid, a setpoint gradient legend, and a transition list — instead of raw attribute JSON — so MSCH support can be exercised before Home Assistant's own integration adds it.
- Adds a Heat/Cool color toggle for Auto-mode schedules: averaging heat+cool setpoints into a single color flattened the gradient into an uninformative blend whenever both bands shared the same center setpoint (found while testing against a real device), so the grid now colors by whichever bound the user selects instead.
- Read-only for this first iteration; `SetActiveScheduleRequest` is intentionally deferred to a follow-up per the issue.

## Test plan
- [x] `npm run format` / `npm run lint` / `npm run build`
- [x] `npm test -w @matter-server/dashboard` — 215/215 passing, including new `ThermostatScheduleUtilTest.ts` coverage of the decode/format/color-mapping logic (tag- and name-keyed wire formats, day-of-week bitmap assignment, midnight wraparound, preset-name resolution, setpoint formatting/clamping)
- [x] Full `npm test` — only pre-existing, unrelated `ws-controller` failure (`ConvertersTest.ts`), reproducible in isolation on code untouched by this branch
- [x] Manually verified against a real MSCH-capable Thermostat device via the local dev server (chips, grid, legend, transitions, Heat/Cool toggle)

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,10 @@ This page shows a detailed overview of the changes between versions without the 
 
 ## **WORK IN PROGRESS**
 
-- Enhancement: (lboue) Dashboard Thermostat cluster view shows a Schedule Configuration panel (MSCH feature) visualizing schedules as a weekly setpoint grid with legend and transition list
 - Enhancement: Introduces Websocket Schema version 13 (backward compatible)
     - (MindFreeze) Adds Websocket command `get_network_topology` and event `network_topology_updated` to expose the Thread and WiFi network details for external visualization
 - Enhancement: (lboue) Dashboard endpoint view shows "Client Clusters" of an endpoint and their binding status when also a Binding cluster is available
+- Enhancement: (lboue) Added a Schedule Configuration panel (Thermostat cluster MSCH feature) to Dashboard visualizing schedules
 - Enhancement: Dashboard dev mode adds a second read button per attribute that reads across all fabrics and shows the result without caching it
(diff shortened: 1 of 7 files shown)


Apollon77:
Follow-up (not in this PR): schedules/transitions the decoder rejects (null ScheduleHandle, Vacation bit set, out-of-range TransitionTime) are silently dropped — the panel can show "No schedules configured." or an empty grid for a device that reported data. Worth a dropped-entry count/diagnostic for the dev/tester audience.

Apollon77 (approved):

lboue:
Addressed the dropped-entry diagnostic follow-up in 5290a6d: `readScheduleDecodeStats()` counts Schedules/transitions rejected by `decodeSchedule`/`decodeTransition` (missing `ScheduleHandle`/`SystemMode`, the reserved Vacation bit, out-of-range `TransitionTime`), surfaced as a small warning badge on the panel when non-zero — so a device with malformed schedule data doesn't silently read as "No schedules configured." Covered by new `readScheduleDecodeStats` unit tests.

PR #976, opened 2026-08-06
Fix EventList attribute wrongly marked writable in dashboard descriptions

agners:
## Summary

The matter.js model misreports the global `EventList` attribute (`0xfffa`) as writable. As a result, every cluster entry in the generated dashboard `descriptions.ts` carried `"writable": true` for `EventList`, making the dashboard offer a write (pencil) action in dev mode that can never succeed — the attribute is server-reported and read-only.

Override it to read-only in `generate-descriptions.ts` and regenerate `descriptions.ts` (flips exactly the 142 `EventList` entries, no other changes).

Noticed via Copilot review feedback on #973, but pre-existing for all clusters.

## Testing

- `npm run format-verify`, `npm run lint`
- Verified the regenerated diff only changes `EventList` (`65530`) entries from `true` to `false`

### DIFF
--- a/packages/dashboard/scripts/generate-descriptions.ts
+++ b/packages/dashboard/scripts/generate-descriptions.ts
@@ -146,7 +146,9 @@ function generateDescriptions(): string {
                     cluster_id: cluster.id,
                     label: toLabel(ace.name),
                     type: getTypeString(ace),
-                    writable: ace.writable,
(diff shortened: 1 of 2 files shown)


Apollon77:
Thanks, I fix it including removal from python. EventList is Deprecated and was basically never allowed because Provisional before Deprecated

PR #978, opened 2026-08-06
Add support for vendor extensions to standard clusters, with WAGO WindowCovering attributes

agners:
## Summary

Some vendors add manufacturer-specific attributes to **standard** Matter clusters (vendor ID in the upper 16 bits of the attribute ID) instead of defining an entirely custom cluster. The custom-clusters package could so far only describe new clusters, not such extensions.

This PR adds a small `clusterExtension()` helper to the custom-clusters package that registers vendor attributes on a standard cluster in the matter.js model. Registered extension attributes are then decoded, typed and exposed like regular attributes — including flowing automatically into the generated Python cluster definitions and the dashboard `descriptions.ts`.

As first user, it describes the Window Covering extension attributes of the **WAGO Home Blind Control** (2757-1105 Wi-Fi / 2757-1106 Thread, vendor ID 0x1534/5428):

| Attribute | ID | Type | Meaning |
|---|---|---|---|
| `WagoTravelTimeUp` | `0x15340001` | uint32, writable | Travel time closed → open, 10 ms units |
| `WagoTravelTimeDown` | `0x15340002` | uint32, writable | Travel time open → closed, 10 ms units |
| `WagoSlatRotationTime` | `0x15340003` | uint32, writable | Full slat rotation time, ms |

Semantics were determined by changing each value in the WAGO Home app and diffing node dumps; all three matched exactly. Note the device also re-measures the travel times on every full end-to-end run (reference run behavior per the product manual), so values may change without an explicit write. A fourth vendor attribute (`0x15340004`, bool) exists but its meaning is not yet identified, so it is deliberately not included.

The generated Python client exposes them as `WindowCovering.Attributes.WagoTravelTimeUp` etc., enabling Home Assistant number entities as follow-up.

## Testing

- `npm run format-verify`, `npm run lint`
- `python_client`: 68 tests pass, including a new test asserting the extension attributes resolve on `WindowCovering`
- Values verified against the real device (dumps with known app settings)

Related: #973


packages/custom-clusters/src/extensions/wago.ts
```diff
@@ -0,0 +1,43 @@
...
+ * the covering. The device re-measures them on every full end-to-end travel,
+ * so their values may change without an explicit write.
+ */
+clusterExtension("WindowCovering", [
+    {
+        id: 0x15340001,
+        name: "WagoTravelTimeUp",
+        type: "uint32",
+        access: "RW VM",
+        conformance: "O",
```
Apollon77: I would keep conformance out here because with this style they can only be "O" ever ... any "M" would create some strange effects potentially ... so lets better hard code "O" in the method above rather than specifying here until we have a need to

Apollon77 (approved):

Issue #1000, opened 2026-08-20
BLE proxy: reference client does not implement write_and_subscribe, so commissioning always fails


ronsilverentand:
### Summary

The BLE proxy server sends `write_and_subscribe`, but the reference proxy client
shipped in the same package never implements it. Commissioning over `--ble-proxy`
therefore cannot complete: the device is discovered and connected, GATT discovery
succeeds, and every attempt then fails at the BTP handshake.

```
[←CMD] id=31 write_and_subscribe {"connection_handle":7,"write_uuid":"18ee2ef5-263d-4559-959f-4f9c429f9d11",
                                  "write_value":"ZWwEAAAAFAD/","write_response":true,
                                  "subscribe_uuid":"18ee2ef5-263d-4559-959f-4f9c429f9d12"}
[→ERR] id=31 Unknown command: write_and_subscribe
```

```
Commission failed: discovery of node with discriminator 3774 failed:
No device could be commissioned (16 of 16 started attempt(s) failed, 16 discovered)
```

### Where the mismatch is

All three files are in `@matter-server/ble-proxy`:

- `dist/esm/ProxyBleChannel.js:135` — the server calls
  `connection.sendCommand(BleProxyCommand.WriteAndSubscribe, {…})` to send the BTP
  handshake on C1 while subscribing C2 atomically. There is no capability check and
  no fallback to the separate `write_characteristic` + `subscribe_characteristic`
  commands.
- `dist/esm/BleProxyProtocol.js:17` — declares
  `WriteAndSubscribe: "write_and_subscribe"`.
- `dist/esm/example/NobleBleProxyClient.js` — `#registerCommandHandlers()` registers
  eleven handlers: `start_scan`, `stop_scan`, `connect`, `disconnect`,
  `discover_services`, `discover_characteristics`, `read_characteristic`,
  `write_characteristic`, `subscribe_characteristic`, `unsubscribe_characteristic`,
  `request_mtu`. `write_and_subscribe` is absent, so `#handleCommand` falls through
  to `Unknown command`.

The handshake protocol reports `protocol v1` from both ends, so there is nothing in
the handshake that would let either side detect the gap.

### Versions

Reproduced on `ghcr.io/matter-js/matterjs-server:1.3.3`
(`matter-server/1.3.3 (matter.js/0.17.7)`) and confirmed still present on `1.4.0` —
same eleven handlers, and `ProxyBleChannel.js:135` unchanged.

### Reproduce

[...]

PR #1005, opened 2026-08-23
fix(ble-proxy): implement write_and_subscribe and tolerate the connect race in NobleBleProxyClient

vavallee:
The Noble reference BLE proxy client cannot complete a BTP session with a real device. Two independent gaps, both hit on every commissioning attempt.

## 1. `write_and_subscribe` is unimplemented

`ProxyBleChannel` sends `BleProxyCommand.WriteAndSubscribe` to write the BTP handshake to C1 and subscribe to C2 atomically (`ProxyBleChannel.ts:198`), but `NobleBleProxyClient` registers no handler for it:

```
[←CMD] id=24 write_and_subscribe {"connection_handle":1,"write_uuid":"18ee...9d11","write_value":"ZWwEAAAA9AD/","write_response":true,"subscribe_uuid":"18ee...9d12"}
[→ERR] id=24 Unknown command: write_and_subscribe
[←CMD] id=25 disconnect {"connection_handle":1}
```

The controller then reports:

```
Unexpected error from parallel commissioning attempt: internal_error: Unknown command: write_and_subscribe
Commission failed: discovery of node with discriminator 15 failed:
  No device could be commissioned (1 of 1 started attempt(s) failed, 1 discovered)
```

From the peripheral's side nothing is ever written — a GATT connect, silence, then a disconnect ~4.5 s later:

```
I (481119) chip[DL]: BLE GAP connection established (con 0)
I (485603) chip[DL]: BLE GAP connection terminated (con 0 reason 0x213)
```

The new handler attaches the data listener **before** the write, because the C2 indication is a separate binary frame that can beat the `write_and_subscribe` response, and `binaryFrameReceived` drops frames emitted with no listener attached — the same ordering constraint `ProxyBleChannel.ts:152-154` documents on the controller side.

## 2. `connect` races the client's own discovery

With that fixed, `connect` still fails intermittently. The controller drives `connect` from its own discovery bookkeeping, which can run ahead of the client's `device_discovered` event for the same advertisement — 104 ms in this capture:

```
23:32:53.443 [←CMD] id=43 connect {"address":"f6:85:47:6b:23:ca"}
23:32:53.443 [CONN] No peripheral found for address "f6:85:47:6b:23:ca". Known:
23:32:53.547 [EVT] device_discovered addr=f6:85:47:6b:23:ca rssi=-54 serviceData=["fff6"]
```

`#handleConnect` failed immediately on an empty `#discoveredPeripherals` map. It now waits up to `CONNECT_DISCOVERY_TIMEOUT_MS` (5 s) for the scan to surface the peripheral before giving up, and the error message says how long it waited.

## Why CI doesn't catch this

`BleProxyIntegrationTest.ts:190` asserts the *controller* emits `write_and_subscribe`, but it drives a `BleProxyTestClient` double. Nothing in the suite exercises `NobleBleProxyClient`, so the reference implementation can drift from the protocol without a single test failing. Happy to add coverage here if you'd like it in the same PR — it wasn't obvious to me how you'd prefer to fake Noble.

## Verification

Against an ESP32-C6 running esp-matter v1.6 (ESP-IDF v5.5.5), commissioned over the `/ble` proxy from matter-server 0.7.1.

Before: every attempt died at the handshake, device silent throughout.

After: BTP session establishes, PASE completes, secure session comes up, 241-byte fragments flow in both directions, and commissioning proceeds to device attestation:

```
[GATT] notify 18ee2ef5263d4559959f4f9c429f9d12 handle=1 241 bytes
[←BIN] opcode=1 handle=1 payload=115 bytes
```

`npx tsc --noEmit -p packages/ble-proxy/tsconfig.json` passes.

Scope note: this is specific to the Node reference client. The Python sibling `matter-ble-proxy` (used by Home Assistant) implements `write_and_subscribe` at `client.py:318` and needs no change.

### DIFF
--- a/packages/ble-proxy/src/example/NobleBleProxyClient.ts
+++ b/packages/ble-proxy/src/example/NobleBleProxyClient.ts
@@ -27,10 +27,18 @@ import {
     type ReadCharacteristicArgs,
     type SubscribeCharacteristicArgs,
     type UnsubscribeCharacteristicArgs,
+    type WriteAndSubscribeArgs,
     type WriteCharacteristicArgs,
     decodeBinaryFrame,
     encodeBinaryFrame,
 } from "../BleProxyProtocol.js";
+
+/**
+ * How long #handleConnect waits for a scan result to surface a peripheral the controller has
+ * already asked us to connect to. The controller drives connect off its own discovery bookkeeping,
+ * which can run ahead of our device_discovered event for the same advertisement.
+ */
+const CONNECT_DISCOVERY_TIMEOUT_MS = 5000;
 
 function withTimeout<T>(promise: Promise<T>, ms: number, message: string): Promise<T> {
     return Promise.race([promise, new Promise<never>((_, reject) => setTimeout(() => reject(new Error(message)), ms))]);
@@ -229,6 +237,9 @@ export class NobleBleProxyClient {
         this.#commandHandlers.set("subscribe_characteristic", (id, args) =>
             this.#handleSubscribeCharacteristic(id, args as unknown as SubscribeCharacteristicArgs),
         );
+        this.#commandHandlers.set("write_and_subscribe", (id, args) =>
+            this.#handleWriteAndSubscribe(id, args as unknown as WriteAndSubscribeArgs),
+        );
         this.#commandHandlers.set("unsubscribe_characteristic", (id, args) =>
             this.#handleUnsubscribeCharacteristic(id, args as unknown as UnsubscribeCharacteristicArgs),
         );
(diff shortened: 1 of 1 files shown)


Apollon77:
Thanks for this — both defects are real, and the ESP32-C6 capture made them easy to confirm. We're taking them in #1003, which was already open on the `write_and_subscribe` gap from #1000, so this lands there rather than as a separate change. Full credit to you for the connect race; I had not found it.

Where #1003 ends up differing from this PR, for your review:

**`write_and_subscribe`** — same write-then-subscribe ordering. #1003 also maps failures to `write_failed` / `subscribe_failed` instead of letting them fall through to `internal_error`, keeps a per-subscription listener reference so `unsubscribe_characteristic` removes its own listener rather than every `"data"` listener, and short-circuits a re-subscribe for an already-subscribed UUID (noble's `_notify` resolves without re-enabling CCCD when already notifying, so a second listener would double every forwarded notification).

One intentional difference: `write_response` defaults to `false` there, not `true`. The protocol doc gives `false` as the default and the Python client uses `args.get("write_response", False)`; the controller always sends `true` explicitly, so it never bites, and matching the spec and the sibling client seemed better than a defensive default that disagrees with both.

Your listener-before-write ordering is equivalent in effect, since the peripheral cannot indicate before CCCD is enabled and both attach before `subscribeAsync()` — which is the window that actually matters, because noble drops `"data"` emitted with no listener attached.

**The connect race** — implemented event-driven rather than polled: a waiter keyed by address, resolved from the `discover` listener. Two reasons beyond taste. It avoids the fixed 100 ms floor when the advertisement is already in flight, and the waiter is woken *before* the duplicate-suppression filter, so with `allow_duplicates: false` a repeat advertisement for a peripheral seen earlier still satisfies a pending connect — the polled version would keep polling in that case.

I also went looking for the mechanism rather than only the symptom, and there is one on the controller side that justifies the guard independently of your capture: `ProxyBleClient`'s discovered-peripheral map is never cleared, and `ProxyBleClient.ts:64` replays cached peripherals to a newly registered discovery callback. So the controller can issue `connect` purely from cache with no fresh advertisement.

**On coverage** — you were right that nothing exercised `NobleBleProxyClient`. #1003 adds a test asserting the handler map covers every `BleProxyCommand`, which is the specific drift that caused #1000, and the same guard on the Python side. It does not fake noble, so the GATT behaviour is still untested; the blocker is that noble arrives through a dynamic `import()` with no injection seam. If you want to add that seam, it would be genuinely useful and is not blocked by anything in #1003.

Also fixed there while in the same file: `JSON.parse` and `decodeBinaryFrame` in the message handler were unguarded, and a throw from either escapes into ws's receiver as an uncaught exception and kills the process — so a single malformed frame ended the bridge mid-commissioning.

#1006 is a separate matter and I'll follow up there.

Apollon77:
@vavallee Thanks for the PR, I contonue with #1003 which was open already for it

PR #1019, opened 2026-08-29
feat(dashboard): add ClosureDimension cluster command panel

lboue:
## Type of change

## Description

Adds a dashboard command panel for the ClosureDimension cluster (0x0105), mirroring the existing ClosureControl panel (which controls a closure as a whole; ClosureDimension controls one degree of freedom/axis of a composed closure, e.g. one panel or one axis of motion).

- Shows **Current**/**Target** state (position %, latch, speed), each gated on the Positioning/MotionLatching/Speed features.
- Shows feature-gated static attributes: Resolution, Step size, Unit + Unit range, Limit range, Translation direction, Rotation axis + Overflow, Modulation type.
- **Set target** form (SetTarget command): position (%), latch (only when LatchControlModes reports remote latching/unlatching support), speed — each field omitted from the payload when left "(unchanged)".
- **Step** form (Step command, Positioning feature only): direction, number of steps, optional speed.

`packages/dashboard/src/util/closure-dimension.ts` holds the cluster helpers (attribute readers, FeatureMap decoding, command wrappers); `packages/dashboard/test/ClosureDimensionUtilTest.ts` covers them (21 cases).

## Backing evidence

- Related issue: #

N/A — this is a new feature (no existing behavior being corrected), not a fix.

## Checklist

- [x] `npm test` passes (dashboard package: 404/404; full monorepo `npm test` only shows the pre-existing, environment-dependent mDNS discovery integration failures noted in `AGENTS.md`, unrelated to this change)
- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This page shows a detailed overview of the changes between versions without the 
 
 ## **WORK IN PROGRESS**
 
+- Enhancement: (lboue) Added a command panel for the ClosureDimension cluster to Dashboard
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
--- a/packages/dashboard/src/pages/cluster-commands/clusters/closure-control-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/closure-control-commands.ts
@@ -123,11 +123,11 @@ class ClosureControlClusterCommands extends BaseClusterCommands {
                     }
 
                     <div class="states-grid">
-                        <div class="state-block">
+                        <div class="state-block state-block--current">
                             <div class="state-block-header">Current</div>
                             ${this._renderState(current, features)}
                         </div>
-                        <div class="state-block">
+                        <div class="state-block state-block--target">
                             <div class="state-block-header">Target</div>
                             ${this._renderState(target, features)}
                         </div>
@@ -379,9 +379,21 @@ class ClosureControlClusterCommands extends BaseClusterCommands {
                 padding: 12px 0;
                 margin-bottom: 12px;
             }
+            .state-block {
+                padding: 8px 12px;
+                border-radius: 8px;
+                background: var(--md-sys-color-surface-container-highest);
+                border-left: 3px solid var(--md-sys-color-outline-variant);
+            }
+            .state-block--target {
+                border-left-color: var(--md-sys-color-primary);
+            }
             .state-block-header {
                 font-weight: 500;
                 margin-bottom: 6px;
+            }
+            .state-block--target .state-block-header {
+                color: var(--md-sys-color-primary);
             }
             .state-fields {
                 display: flex;
--- a/packages/dashboard/src/pages/cluster-commands/clusters/closure-dimension-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/closure-dimension-commands.ts
@@ -1,0 +1,455 @@
+/**
+ * @license
+ * Copyright 2025-2026 Open Home Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import "@material/web/button/filled-button";
+import "@material/web/button/outlined-button";
+import "@material/web/select/outlined-select";
+import "@material/web/select/select-option";
+import { css, html, nothing } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import { handleAsync } from "../../../util/async-handler.js";
+import {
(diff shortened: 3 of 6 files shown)


lboue:
**Testing**

image

Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`).

**Copilot's `NumberOfSteps` comment is correct**, but it is one symptom of a shape problem rather than a missing `max`: the Step field was coerced on every keystroke (`parseInt` → `NaN`/`0` → `1`), so clearing it snapped it back to `1` with the caret at the end and you could never type a shorter value; `parseInt("1e5")` silently became `1`; and anything above 65535 reached the device and was rejected with no visible result. The Position field, by contrast, was already validated at submit time.

What the commit does:

- Moves both checks into `parseTargetPositionPercent()` / `parseNumberOfSteps()` in `util/closure-dimension.ts`, so the panel keeps the raw string and validity is decided when Set/Step is pressed. Both buttons disable while a field is unsubmittable. Unit-tested, including the 65535/65536 boundary.
- Position is now also bounded by `LimitRange` when the Limitation feature is present. The panel already *displayed* that range but still let you submit 100% on a 0–80% dimension, which the device rejects.
- Command failures now go through `showAlertDialog` like the door-lock/chime/thermostat/AVSUM panels do. Previously `handleAsync` had no `onError`, so every rejection was console-only — that was copied from the ClosureControl panel, but ClosureControl has only enum selects and no user-reachable out-of-range path, which is what made silent failure newly load-bearing here.
- The "Set target" block is omitted when no control inside it is available (an `LT`-only dimension with `LatchControlModes = 0` rendered a header and a permanently disabled button).

Not changed, flagged for you: the PR also restyles the existing ClosureControl state blocks (new container background, primary-coloured target accent). The tokens are theme-correct, but it is not mentioned in the PR body or the CHANGELOG — worth a line in the description.

Gates: `npm run format`, `npm run lint`, `npm run build`, `npm test -w @matter-server/dashboard` all green.

Apollon77:
@lboue one thing for the PR description before this merges — the code is fine, this is only about what the description claims.

The PR also **restyles the existing ClosureControl panel** (`closure-control-commands.ts`): the state blocks gain a `--md-sys-color-surface-container-highest` background and the target block a primary-coloured left border and header. The tokens are correct and resolve in both themes, so there is nothing to fix — but neither the description nor the CHANGELOG mentions that an already-shipped panel changes appearance, and a reviewer diffing the two files would not expect it.

Suggested additions:

- **Description** — a line under the bullet list: *"Also aligns the existing ClosureControl panel's state-block styling with the new panel, so the two read as one family."*
- **CHANGELOG** — the entry currently only announces the new panel. Either extend it or add a second line, e.g. `- Enhancement: (lboue) Dashboard ClosureControl panel's state blocks use the same styling as the new ClosureDimension panel`.

Nothing else in the description needs changing — I checked every functional claim against the code, including the "21 cases" count, and they all hold.

Apollon77:

Both `handleAsync` error callbacks could run after the panel had been reused for another node, so a SetTarget or Step rejection from the previous device would open its dialog over the current one. `_failureReporter(command)` now captures `node`/`endpoint` at render time and checks `isSameContext` before showing the dialog, logging instead when the panel has moved on — same shape as `AGENTS.md` requires and as I used on #1023.

That gap was mine, not lboue's: the dialogs did not exist before my previous commit, which is exactly the case the rule exists for (a fix that makes a latent path reachable). 433/433 green.

Apollon77:
Correct — fixed and pushed.

`Math.round(percent * 100)` quietly turned a typed `12.345%` into `12.35%`. The button's disabled state is driven by this parser rather than the text field's native `step` validity, so a value the command cannot represent has to be rejected here rather than rounded into something the operator did not ask for. `parseTargetPositionPercent` now returns null unless the value is an integral number of hundredths; `12.34` still parses, `12.345` and `0.001` do not.

411/411 on a clean build.

Apollon77:

Collecting the description-level corrections from the review rounds here, so they are in one place rather than spread over several comments. The PR body itself is unchanged; treat this as part of it.

**Also changes an existing panel.** Besides adding the ClosureDimension panel, this PR restyles the existing **ClosureControl** state blocks — new container background, primary-coloured accent on the target block. The tokens resolve in both themes, so there is nothing to fix, but the appearance of an already-shipped panel changes and neither the description nor the CHANGELOG says so.

**Validation is stricter than the original description implies.** `SetTarget`/`Step` input is validated when the button is pressed, not coerced per keystroke:

- Position is bounded by `LimitRange` when the Limitation feature is present, not just 0–100 %.
- Position finer than a hundredth of a percent is rejected rather than rounded — `12.34` is sent, `12.345` is refused, because `percent100ths` cannot carry it.
- `NumberOfSteps` is bounded to the uint16 range, and the field can be cleared.
- Both buttons disable while a field holds no submittable value.

**Command failures are now visible.** A device rejection or transport failure raises `showAlertDialog`, matching the door-lock/chime/thermostat panels; previously it was `console.error` only. The dialog is guarded by panel context, so a rejection from a node the panel has since navigated away from is logged instead of shown.

**"Set target" is omitted entirely** when no control inside it is available — an `LT`-only dimension with `LatchControlModes = 0` used to render a header and a permanently disabled button.

Apollon77 (approved):

PR #1022, opened 2026-08-30
feat(dashboard): show per-endpoint label in Endpoints/Clusters panels

lboue:
## Type of change

## Description

Devices with many endpoints (bridges in particular) show only "Endpoint N" in the
dashboard's Endpoints list and Clusters-on-endpoint panel, with no way to tell what
each endpoint represents without opening it. This adds two complementary,
independent ways to identify an endpoint at a glance:

**1. Per-endpoint label** — `getEndpointLabel()` resolves a human-readable label,
in priority order:
1. `BridgedDeviceBasicInformation` → `NodeLabel` (57/5) — the per-endpoint NodeLabel
   exposed for bridged devices (e.g. Matterbridge).
2. `UserLabel` → `LabelList` (65/0)
3. `FixedLabel` → `LabelList` (64/0)

(BasicInformation's own `NodeLabel` is endpoint-0-only and describes the whole node,
so it's intentionally not part of this resolution.)

**2. Semantic tag chips** — `getEndpointSemanticTags()` decodes the Descriptor
cluster's `TagList` (29/4) for the endpoint and renders each tag as a chip (e.g.
"Common Area Namespace → Living Room"), reusing the same `.chip`/`.chip-list` style
already used for the Descriptor cluster's own "Semantic Tags" panel (now extracted
as `chipListStyles` for reuse).

Both are rendered in the Endpoints list (`matter-node-view.ts`) and the
Clusters-on-Endpoint panel (`matter-endpoint-view.ts`, header included); the label
is also shown in the cluster detail header (`matter-cluster-view.ts`).

## Backing evidence

Not applicable — this is an additive feature (new label/tag resolution + display),
not a fix for existing wrong behavior.

- Related issue: #

## Checklist

- [x] `npm test` passes — full `npm test` reports only the 2 pre-existing "Device Discovery" integration
      failures documented in `CLAUDE.md` (mDNS-dependent, environment-specific, unrelated to this change).
      Scoped `npm test -w @matter-server/dashboard` passes 414/414, including the new tests.
- [x] CHANGELOG updated


lboue:
**Testing**

image

Apollon77:
@lboue can you look into the comments please?

Apollon77:
Reviewed and pushed a fix commit on top of your latest push.

**The three open Copilot comments were already fixed** by your `d60d069c`/`66fe7933` — both new renderers preserve `erroneous`, and `.chip.chip-error` is in `chipListStyles`. I have not re-done them; they can be resolved.

What the new commit adds:

- **NUL sanitation was only on the bridged path.** You had just fixed `BridgedDeviceBasicInformation.NodeLabel`, but `decodeLabelListValues` had no equivalent, so a bridge that pads a fixed-width `UserLabel` value renders `Kitchen\0\0\0` into the endpoints list and the header title. All three sources now go through one `labelText()` helper.
- It is built on the existing `tagField`/`toText` helpers from `attribute-shapes.ts` instead of a local `isRecord` + `entry["1"]` + trim — those helpers exist for exactly this tag-keyed struct decoding (and `isRecord` also returned true for arrays).
- **The joined label was unbounded.** `UserLabel` is a list; 8 entries of 16 chars produce a ~150-char parenthetical that pushes `| Cluster N (Name)` off a single-line header. Capped at 3 entries / 60 chars with an ellipsis, matching `semantic-tags.ts`'s own `RAW_ENTRY_TEXT_MAX = 60`.
- **The chip markup existed in three copies** and the two new ones each had to be corrected separately when they dropped the `erroneous` flag — that is the regression class, not the individual sites. It now renders from one `renderSemanticTagChips()` in `components/semantic-tag-chips.ts`, used by the node view, the endpoint view and the Descriptor detail panel.
- Related: the same tag rendered at 0.75rem in the Endpoints list and 0.85rem one click later in the Clusters panel, because only `matter-node-view` carried the compact override. That is now a `.chip-compact` modifier in `chipListStyles`, applied in both.
- Tests: NUL padding on both paths, blank-`UserLabel` fall-through to `FixedLabel`, the index-keyed (non-array) LabelList shape, the entry cap and the ellipsis.

Two things for the description, not the code: the checklist says "414/414" but the dashboard package has nothing near that count, and `- Related issue: #` is still an unfilled placeholder.

Apollon77:
@lboue two description-only corrections, no code change needed for either.

**1. The test count is wrong.** The checklist says *"Scoped `npm test -w @matter-server/dashboard` passes 414/414"*. The dashboard package has nothing near that many tests — it is 383 on `main` and 393 on this branch before my commit. 414 corresponds to no run of this package I can reproduce; most likely it came from a working tree that still had compiled test artifacts from another branch in `dist/` (that trap is real — a deleted `test/*.ts` leaves its `dist/esm/test/*.js` behind and it keeps running, gitignored so invisible). Worth re-running after a `npm run build-clean` and quoting the real number.

**2. `- Related issue: #` is still an unfilled placeholder** — either link the issue or drop the line.

Also worth adding, since it is a genuine behaviour of the feature that the description does not mention: **`getEndpointLabel()` now rejects NUL-padded values from all three sources**, and the joined label is capped at 3 entries / 60 characters so a `UserLabel` list cannot push the cluster name out of the header. Both are in the commit I pushed; a reader of the description would otherwise be surprised that a long label is truncated.

Apollon77:
Correct — removed. The doc comment on `chipListStyles` restated the export name and the selectors directly below it, which is the WHAT-restatement the repo's comment rule says to delete on sight. It was mine, added when I extracted the fragment.

430/430 green. No other findings this round.

Apollon77:

**Correction — the test count.** The checklist says *"Scoped `npm test -w @matter-server/dashboard` passes 414/414"*. That number matches no run of this package: it is 383 on `main` and 430 on this branch as it now stands. 414 most likely came from a working tree carrying compiled test artifacts from another branch in `dist/` — a deleted `test/*.ts` leaves its `dist/esm/test/*.js` behind and it keeps running, gitignored so invisible. I hit exactly this while reviewing and had to `rm -rf packages/dashboard/dist` to get honest numbers. **430/430** after a clean build.

**`- Related issue: #` is an unfilled placeholder** in the body.

**Two behaviours the description does not mention**, both added during review:

- `getEndpointLabel()` rejects NUL-padded values from all three sources, not only the bridged `NodeLabel` — a bridge padding a fixed-width `UserLabel` would otherwise render `Kitchen\0\0\0` into the endpoints list.
- The joined label is capped at 3 entries / 60 characters with an ellipsis, so a device exposing many `UserLabel` entries cannot push the cluster name out of a single-line header.

**One refactor worth knowing about:** the TagList chip markup existed in three copies, and the two new ones each had to be corrected separately when they dropped the decoder's `erroneous` flag. It now renders from one `renderSemanticTagChips()` used by the node view, the endpoint view and the Descriptor detail panel, with the compact variant as a `chipListStyles` modifier — so the same tag no longer changes size between the endpoints list and the clusters panel.

PR #1023, opened 2026-08-30
feat(dashboard): add a command panel for the MediaPlayback cluster

lboue:
## Type of change

## Description

Adds a dedicated command panel for the MediaPlayback cluster (ID 1286 / 0x0506) to the Dashboard, replacing the generic "DEV Commands" list previously shown for that cluster.

- A playback state chip (Playing / Paused / Not playing / Buffering), decoded from `CurrentState`.
- A position/duration readout (`SampledPosition` / `Duration`), formatted as `m:ss` (or `h:mm:ss`), plus the current `PlaybackSpeed` when it differs from 1x.
- A transport control row: Previous, Rewind, Play, Pause, Stop, Fast Forward, Next, Start Over.
- Skip Forward / Skip Backward with a configurable delta (ms).

Every control is shown only if the device actually advertises the corresponding command in its `AcceptedCommandList`, matching the pattern already used for the DoorLock panel. `Seek`, `ActivateAudioTrack`, `ActivateTextTrack`, and `DeactivateTextTrack` are intentionally left out of this first pass (track selection / seek position UI is a separate, larger piece of work).

### Screenshot

image

## Backing evidence

Not applicable — this is a new feature, not a fix for existing wrong behavior.

- Related issue: #

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This page shows a detailed overview of the changes between versions without the 
 
 ## **WORK IN PROGRESS**
 
+- Enhancement: (lboue) Added a command panel for the MediaPlayback cluster to Dashboard, with transport controls (Play, Pause, Stop, Previous, Next, Rewind, Fast Forward, Start Over, Skip Forward/Backward) gated by the device's AcceptedCommandList, plus a playback state/position readout
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
--- a/packages/dashboard/src/pages/cluster-commands/clusters/media-playback-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/media-playback-commands.ts
@@ -1,0 +1,252 @@
+/**
+ * @license
+ * Copyright 2025-2026 Open Home Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import "@material/web/button/outlined-button";
+import "@material/web/iconbutton/outlined-icon-button";
+import {
+    mdiFastForward,
+    mdiPause,
+    mdiPlay,
+    mdiRestart,
+    mdiRewind,
+    mdiSkipBackward,
+    mdiSkipForward,
+    mdiSkipNext,
+    mdiSkipPrevious,
+    mdiStop,
+} from "@mdi/js";
+import { css, html, nothing } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import "../../../components/ha-svg-icon.js";
+import { handleAsync } from "../../../util/async-handler.js";
+import {
+    FAST_FORWARD_COMMAND_ID,
+    formatDurationMs,
+    formatPlaybackState,
+    MEDIA_PLAYBACK_CLUSTER_ID,
+    NEXT_COMMAND_ID,
+    PAUSE_COMMAND_ID,
+    PlaybackState,
+    PLAY_COMMAND_ID,
+    PREVIOUS_COMMAND_ID,
+    readCurrentState,
+    readDurationMs,
+    readPlaybackSpeed,
+    readPositionMs,
+    REWIND_COMMAND_ID,
+    SKIP_BACKWARD_COMMAND_ID,
+    SKIP_FORWARD_COMMAND_ID,
+    START_OVER_COMMAND_ID,
+    STOP_COMMAND_ID,
+    supportsCommand,
+} from "../../../util/media-playback.js";
+import { BaseClusterCommands } from "../base-cluster-commands.js";
+import { registerClusterCommands } from "../registry.js";
+
+const DEFAULT_SKIP_MS = 10000;
+
+@customElement("media-playback-cluster-commands")
+class MediaPlaybackClusterCommands extends BaseClusterCommands {
+    @state() private _skipMs = DEFAULT_SKIP_MS;
+
+    override render() {
+        if (!this.node || this.cluster !== MEDIA_PLAYBACK_CLUSTER_ID) return nothing;
+
+        const playbackState = readCurrentState(this.node, this.endpoint);
+        const positionMs = readPositionMs(this.node, this.endpoint);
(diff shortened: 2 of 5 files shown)


Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`). Cluster IDs, command IDs, the `PlaybackPositionStruct` tags and the `deltaPositionMilliseconds` field name all check out against `@matter/model`, and gating on `AcceptedCommandList` rather than the feature bits is the stricter and correct choice.

The substantive finding was that **command failures were invisible twice over**:

1. `BaseClusterCommands.sendCommand` catches and returns normally, so the `handleAsync` error paths on every transport button were unreachable — a transport failure was a silent no-op.
2. More importantly, MediaPlayback answers with `PlaybackResponse{Status}`, and a non-Success status is a *successful* invoke at the interaction layer. Pressing Rewind on a device already at maximum rewind speed returns `SpeedOutOfRange(4)`; the payload was discarded, so nothing happened and nothing was said. Same for `InvalidStateForCommand` on Play while buffering.

`invokeTransportCommand()` now decodes the status (name-keyed — command responses go through `convertMatterToWebSocketNameBased`, unlike attributes, so `pickNumber(obj, "status")` is right and a tag lookup would not be) and raises it through `showAlertDialog`, which is what 8 of the 13 sibling panels do. The dialog captures node/endpoint at render time and checks `isSameContext` before opening, per `cluster-commands/AGENTS.md`.

Also in the commit:

- **The skip field could not be edited by deleting.** `_handleSkipMsChange` wrote `String(this._skipMs)` back on every `input` event, so the keystroke that empties the box reset it to 10000 with the caret at the end. It now keeps the raw string and validates on submit (`parseSkipMs`), with the buttons disabled while it is unsubmittable, and it resets when the panel is reused for another node.
- **`PlaybackSpeed` is a float32**, so a device reporting 0.1x rendered `0.10000000149011612×`; it is rounded now. A stopped AS-capable device reports `0`, which passed the `!== 1` guard and displayed a meaningless `0×`.
- **`renderWhenOffline` was missing**, so `matter-cluster-view` unmounted the panel entirely while the node was unreachable and every `?disabled=${!online}` was dead code. The panel decodes cached attributes, which is exactly what the flag documents, so it is registered with it and the offline gating now does real work.
- When `AcceptedCommandList` is not in the cache the panel collapsed to a bare state chip with no explanation; it now says so.
- Tests for the null `Position` live-stream shape, bigint position, `Duration = 0` vs absent, the 59:59/1:00:00 hour boundary, `parseSkipMs`, and the status decoding. I mutation-tested the two new guards (inverted each; three tests fail, restored).
- `const enum` → `enum` (it was the only `const enum` in the package) and the `mm:ss` JSDoc corrected to `m:ss`.

One description correction: "replacing the generic 'DEV Commands' list previously shown for that cluster" is not accurate — `matter-cluster-view.ts` renders the dev-commands panel whenever dev mode is on, with no registry check, so both panels appear for cluster 1286.

Apollon77:
@lboue one description correction, and one addition now that the panel behaves differently.

**1. The panel does not replace the DEV Commands list.** The description opens with *"replacing the generic 'DEV Commands' list previously shown for that cluster"*. `matter-cluster-view.ts` renders `_renderDevCommandsPanel()` whenever dev mode is on, with no check against the cluster-commands registry, and this PR does not touch that. So on cluster 1286 **both** panels appear when dev mode is enabled. Suggest: *"Adds a dedicated command panel for the MediaPlayback cluster (ID 1286 / 0x0506) to the Dashboard, alongside the generic DEV Commands list."*

**2. "matching the pattern already used for the DoorLock panel"** overstates it — `door-lock-commands.ts` calls `supportsCommand` exactly once, for `UnlockWithTimeout`; it does not gate its controls generally. Gating *every* control on `AcceptedCommandList` is new here, and it is the right call, so it is worth claiming as its own decision rather than as an existing pattern.

**3. Worth adding, from the commit I pushed:** the panel now **stays visible while the node is unreachable** (`renderWhenOffline`), so the cached state/position readout survives a device dropping off the network while the controls grey out — before, the whole panel was unmounted. And a **device-side rejection is now surfaced**: MediaPlayback answers with `PlaybackResponse{Status}`, and a non-Success status is a successful invoke at the interaction layer, so e.g. Rewind at maximum speed used to do nothing silently. Both change what a user sees, so they belong in the description.

The screenshot still matches. Everything else in the description checks out, including the `m:ss` / `h:mm:ss` formatting and the deliberate exclusion of Seek and the track commands.

Apollon77:

The message named an unread `AcceptedCommandList` as the only cause, but a device that reports a list containing only `Seek` or the track-selection commands (all four deliberately out of scope for this panel) also advertises nothing this panel renders, and was told its list had not arrived. New wording covers both:

> This device advertises none of the transport commands this panel offers. Its AcceptedCommandList may not have been read yet, or it may support only commands handled elsewhere, such as Seek or track selection.

Gates green: format, lint, build, `npm test -w @matter-server/dashboard` 406/406.

Apollon77:

**Correction — it does not replace the DEV Commands list.** The description opens with *"replacing the generic 'DEV Commands' list previously shown for that cluster"*. `matter-cluster-view.ts` renders `_renderDevCommandsPanel()` whenever dev mode is on, with no check against the cluster-commands registry, and this PR does not touch that. On cluster 1286 with dev mode enabled, **both** panels appear. Read that line as "adds a dedicated panel alongside the generic DEV Commands list".

**Correction — the DoorLock comparison is overstated.** *"matching the pattern already used for the DoorLock panel"*: `door-lock-commands.ts` calls `supportsCommand` exactly once, for `UnlockWithTimeout`. Gating *every* control on `AcceptedCommandList` is new here — it is the right call, just not an existing pattern being followed.

**Two behaviours added during review that change what a user sees:**

- **Device rejections are surfaced.** MediaPlayback answers with `PlaybackResponse{Status}`, and a non-Success status is a *successful* invoke at the interaction layer — so Rewind on a device already at maximum speed, or Play while buffering, previously did nothing and said nothing. The status is decoded (name-keyed, as command responses are) and raised through `showAlertDialog`.
- **The panel stays mounted while the node is unreachable** (`renderWhenOffline`), so the cached state/position readout survives a device dropping off the network while the controls grey out. Before, the whole panel was unmounted and the `online` gating was unreachable code.

**Smaller:** `PlaybackSpeed` is rounded away from its float32 representation (`0.10000000149011612×` → `0.1×`) and a stopped player's `0` no longer renders as `0×`; the skip field can be cleared and is validated on submit; and when the device advertises none of this panel's transport commands, it says so instead of showing an empty row.

The screenshot still matches.

Issue #1025, opened 2026-09-01
Add support for expiring PINs in the Door Lock dashboard


lboue:
### Description

The Door Lock dashboard currently provides user and credential management, as well as WDSCH, YDSCH and HDSCH access schedules.

However, Matter also defines an `ExpiringUser` user type specifically for users whose access expires after a configurable timeout. This makes it possible to create temporary PIN credentials without requiring a YDSCH schedule.

The current Door Lock dashboard does not expose this capability.

### Describe the solution you'd like

Add support for creating and managing **expiring users / temporary PINs** in the Door Lock dashboard.

For a Door Lock supporting the `USR` feature, the dashboard should allow a user to be created with:

* `UserType = ExpiringUser` (`7`)
* a PIN credential
* an expiration timeout using the `ExpiringUserTimeout` attribute

The UI could expose this as a **"Temporary PIN"** or **"Expiring PIN"** option when creating a Door Lock user.

For example:

```text
User
  Name: Guest
  Type: Expiring user

PIN
  123456

Expiration
  24 hours
```

The dashboard should also display the user type and expiration information for existing expiring users.

### Matter specification

The Matter Door Lock cluster defines `ExpiringUser` as a value of `UserTypeEnum`:

```text
UserTypeEnum.ExpiringUser = 7
```

It is optionally supported with the `USR` feature.

The cluster also defines the `ExpiringUserTimeout` attribute, which is used to configure the expiration timeout for an expiring user.

This is distinct from `YDSCH`:

* `YDSCH` provides date/time access schedules.
* `ExpiringUser` provides an expiration-based user type.

### Feature detection

The dashboard should detect this capability based on the Door Lock cluster capabilities rather than introducing a new `FeatureMap` bit for `ExpiringUser`.

In particular:

```text
FeatureMap.USR
+
ExpiringUser / ExpiringUserTimeout support
```

`ExpiringUser` itself is **not** a `FeatureMap` feature.

### Describe alternatives you've considered

A temporary PIN could also be implemented using a `ScheduleRestrictedUser` together with a YDSCH schedule.

However, this is a different Matter mechanism and is more appropriate when access needs to be restricted to specific date/time windows.

For a simple **"PIN expires after X minutes/hours"** use case, `ExpiringUser` is the native Matter mechanism.

### Additional context

This would complement the Door Lock user and credential management introduced in PR #1004:

https://github.com/matter-js/matterjs-server/pull/1004

[...]

PR #1031, opened 2026-09-02
feat(dashboard): support expiring/temporary PIN users on Door Lock

lboue:
## Type of change

## Description

Adds support for temporary/expiring PIN users to the Door Lock cluster panel, detected via the
existing `USR`/`PIN` feature bits rather than a new feature flag:

- `addUser()` now accepts `userType`/`userStatus` instead of always creating an `UnrestrictedUser`
  record.
- New `attachPinCredential()` finds a free PIN credential slot (walking `GetCredentialStatus` the
  same way `readUsers` already walks `GetUser`) and attaches a PIN to an existing user via
  `SetCredential`.
- The "Add user" editor gained a "Temporary PIN" type (shown only when the lock reports the `PIN`
  feature), which creates the user and its credential in one action.
- `ExpiringUserTimeout` is now readable and writable from a lock-wide control in the panel.
- User badges now show whether a PIN credential is set, and the expiry hint for `ExpiringUser`
  records.

Out of scope for this PR (flagged for a possible follow-up): converting an existing non-expiring
user into an expiring one after the fact, editing/removing a PIN on an existing user, and
RFID/fingerprint/face credentials.

## Backing evidence

- Related issue: #1025
- [ ] This fix/change is backed by a **complete log file** — N/A, this is a feature addition, not a
      fix for observed wrong behavior.

## Checklist

- [x] `npm test` passes (scoped `npm test -w @matter-server/dashboard`: 435/435; the full `npm test`
      only fails on the two pre-existing mDNS `IntegrationTest` "Device Discovery" cases documented
      as environment-dependent in `CLAUDE.md`, unrelated to this change)
- [x] CHANGELOG updated

---

Opened as a draft: this was scaffolded by Claude Code from issue #1025 and needs human review
before it's ready — see the AI policy link above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ This page shows a detailed overview of the changes between versions without the 
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
 - Enhancement: (lboue) DoorLock cluster panel supports creating temporary/expiring PIN users (UserType=ExpiringUser) and configuring the lock's ExpiringUserTimeout
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
+- Fix: Door Lock PIN fields (`credentialData`, `pinCode`) are redacted from the debug logs of both the server request path and the ws-client send path, matched independently of the key's casing; ws-client exports `redactSensitiveCommandFields()` for the same purpose
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
 
 ## 1.4.0 (2026-08-07)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This page shows a detailed overview of the changes between versions without the 
 ## **WORK IN PROGRESS**
 
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
+- Enhancement: (lboue) DoorLock cluster panel supports creating temporary/expiring PIN users (UserType=ExpiringUser) and configuring the lock's ExpiringUserTimeout
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
 
--- a/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
@@ -762,9 +762,10 @@ class DoorLockClusterCommands extends BaseClusterCommands {
             this._userEditorError = nameError;
             return;
         }
-        const expiring = this._newUserType === USER_TYPE_EXPIRING;
+        const canAddExpiring =
+            isFeatureActive(node, endpoint, "PIN") && isFeatureActive(node, endpoint, "USR");
+        const expiring = canAddExpiring && this._newUserType === USER_TYPE_EXPIRING;
         const pin = this._newUserPin.trim();
-        if (expiring) {
             const pinError = pinCodeLengthError(
                 pin,
                 readMinPinCodeLength(node, endpoint),
--- a/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
@@ -762,10 +762,10 @@ class DoorLockClusterCommands extends BaseClusterCommands {
             this._userEditorError = nameError;
             return;
         }
-        const canAddExpiring =
-            isFeatureActive(node, endpoint, "PIN") && isFeatureActive(node, endpoint, "USR");
+        const canAddExpiring = isFeatureActive(node, endpoint, "PIN") && isFeatureActive(node, endpoint, "USR");
         const expiring = canAddExpiring && this._newUserType === USER_TYPE_EXPIRING;
         const pin = this._newUserPin.trim();
+        if (expiring) {
             const pinError = pinCodeLengthError(
                 pin,
                 readMinPinCodeLength(node, endpoint),
--- a/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/door-lock-commands.ts
@@ -31,6 +31,7 @@ import "../../../components/ha-svg-icon.js";
 import { handleAsync } from "../../../util/async-handler.js";
 import {
     addUser,
+    attachPinCredential,
     buildDaySegments,
     clearHolidaySchedule,
     clearWeekDaySchedule,
@@ -40,12 +41,14 @@ import {
     DOOR_LOCK_CLUSTER_ID,
     formatDaysMask,
     formatDoorState,
+    formatExpiringTimeoutHint,
     formatScheduleStatus,
     formatLockState,
     formatLockType,
     formatOperatingMode,
     formatWallClock,
     defaultHolidayMode,
+    hasPinCredential,
     holidayModeChoices,
     formatTimeOfDay,
     formatUserLabel,
@@ -60,12 +63,17 @@ import {
     nextFreeUserIndex,
     nowAsWallClock,
     parseTimeOfDay,
+    pinCodeLengthError,
     readActuatorEnabled,
     readDoorState,
+    readExpiringUserTimeout,
     readHolidaySchedule,
     readHolidaySchedulesSupported,
     readLockState,
     readLockType,
+    readMaxPinCodeLength,
+    readMinPinCodeLength,
+    readNumberOfPinUsersSupported,
     readTotalUsersSupported,
     readUser,
     readUsers,
@@ -85,7 +93,10 @@ import {
     unlockWithTimeout,
     userNameLengthError,
     USER_NAME_MAX_LENGTH,
+    USER_STATUS_OCCUPIED_ENABLED,
+    USER_TYPE_EXPIRING,
     weekDayScheduleRangeError,
+    writeExpiringUserTimeout,
     writeHolidaySchedule,
     writeWeekDaySchedule,
     writeYearDaySchedule,
@@ -111,6 +122,12 @@ const DATE_TIME_MAX = "2136-02-07T06:28:15";
 
 /** Bounds the GetUser walk on a lock that leaves NumberOfTotalUsersSupported unreported. */
 const USER_SCAN_FALLBACK = 32;
+
+/** Bounds the free PIN credential slot search on a lock that leaves NumberOfPinUsersSupported unreported. */
+const PIN_CREDENTIAL_SCAN_FALLBACK = 32;
+
+/** UserTypeEnum.UnrestrictedUser — the default "Standard" choice in the add-user editor. */
+const USER_TYPE_STANDARD = 0;
 
 interface WeekDayEditor {
     kind: "weekDay";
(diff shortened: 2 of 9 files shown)


lboue:
Testing

image

Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`). Everything the PR asserts about the cluster checks out against `@matter/model` — `ExpiringUser = 7`, `ExpiringUserTimeout` id 53 with constraint `1 to 2880`, `CredentialTypeEnum.Pin = 1`, the `Duplicate`/`Occupied` status codes, and base64 octstr on the wire.

**The redaction had a hole that mattered.** `SENSITIVE_PAYLOAD_FIELDS` matched `credentialData`/`pinCode` with an exact-case `key in payloadRecord`, but the server deliberately accepts non-camelCase payload keys — `Converters.ts:154` runs `camelize()` over every member, and the repo's own converter test uses `PINCode:`, which is the python-matter-server / Home Assistant spelling. So an HA-originated `LockDoor {PINCode: "..."}` was logged verbatim with the base64 PIN, while the dashboard — the one client using the exact camelCase spelling — was the only one redacted. Matching is case-insensitive now, with tests for `PINCode` and `CredentialData`.

Copilot's malformed-JSON comment is mechanically correct but lower priority than that (nothing real emits malformed JSON, and a client crafting one is logging its own PIN). I fixed it anyway since it is cheap: an unparseable frame now logs `<unparseable request, N bytes>` instead of its contents. That also let me widen `redactSensitiveCommandFields` to take an `object`, which removes the `as { args?: unknown }` cast at the call site.

Also in the commit:

- **`nextFreePinCredentialIndex()` trusted an optional response field.** `GetCredentialStatusResponse.NextCredentialIndex` is conformance `O`; a lock that omits it ends the chain after index 1, so the candidate scan handed back index 2 whether or not it was occupied. `SetCredential` then fails with `Occupied(3)` *after* `SetUser` has already created the user, leaving an expiring user with no PIN and no retry path. Any index the chain did not reach is probed directly now. Test covers the omitting lock, and the existing out-of-range-chain test was updated (it now expects the extra probe).
- **The success path after `attachPinCredential` had no `isSameContext` guard.** The failure path had one, but the success path fell through to `_selectedUserIndex = userIndex` and cleared `_weekDaySlots`/`_yearDaySlots` — several round trips after the last check. Navigating to another lock mid-attach stamped lock A's user index onto lock B and wiped B's schedule state. `cluster-commands/AGENTS.md` requires the guard before the write.
- The same `catch` dropped the error entirely when the panel had moved, so a created-user-without-PIN could go completely unrecorded. It is logged now, matching `#reportFailure`.
- Removed the comment at `#renderAddUserEditor()` — the parenthetical about `SetCredential`'s combined-creation use case is the rejected-alternative pattern `CLAUDE.md` says to delete, and the trailing clause restates what `#expiringUserSupported()`'s own jsdoc says. Copilot was right about that one. Also corrected the `ExpiringUserTimeout` spec reference (the constraint is §5.2.9.36, not §7.18.2.51 — `door-lock.ts:314` already cites it correctly).
- CHANGELOG: the entry covered only the dashboard, but this PR also adds a new public export to the published `@matter-server/ws-client` package and changes server log output. Added a line for it; the PR description should mention it too, and the commit scope is `feat(dashboard)` for a diff touching three packages.

Three things I did **not** change, for your call:

1. **`this._newUserPin.trim()`.** A PIN is an octstr and spaces are legal bytes, so trimming silently programs a different secret than the operator typed. But `#pinForCommand()` trims the unlock PIN too, so the two are currently consistent — fixing only the new one would mean a PIN programmed with whitespace could never be used to unlock. Either both or neither; I left it alone rather than split them.
2. **`SetUser` → `SetCredential` is non-atomic** and the failure leaves an orphan user that looks live in the badge row but can never open the door, with no in-panel way to fix it. Inverting the order (`SetCredential` creates user + credential + type, then `SetUser` for the name) turns the failure mode into "working PIN user with no name", which is strictly better and needs no cleanup path. That is a design change, so it is yours to decide.
3. **Bigger adjacent exposure, out of scope here:** `set_wifi_credentials` (`args.credentials`), `set_thread_dataset` (`args.dataset`) and the DoorLock Aliro keys are top-level secrets logged verbatim by both `Connection.sendMessage` and the server request log. `redactSensitiveCommandFields` is named generically but only inspects `args.payload`. Worth a follow-up.

Coverage gap worth noting: no test asserts that `addUser` sends `userType: 7, userStatus: 1` — the PR's headline behaviour change. A regression to `null, null` passes the suite today.

Gates: format, lint, build, `npm test -w @matter-server/dashboard` and `npm test -w @matter-server/ws-client` all green.

Apollon77:
Follow-up to my review comment — two of the three items I had left open are now done, pushed to this branch.

**1. The orphan window is closed.** `SetUser` → `SetCredential` is still two commands, but the sequence now lives in `createExpiringPinUser()` in `util/door-lock.ts` and undoes itself:

- PIN attached → nothing changes for the caller.
- PIN failed, `ClearUser` succeeded → the lock is back to where it started, so the editor **keeps the entered name and PIN** and shows the reason inline: *"… The user was removed again; nothing was changed."* A retry is then safe and lands on the same free index.
- PIN failed **and** `ClearUser` failed → the user is stranded, so the editor closes and an alert names exactly what is on the lock and what to do: *"… User 4 ("Guest") was created on the lock but has no PIN, and removing it again also failed (…). It cannot open the door — delete it from the user list and try again."*

Moving the sequence into the util is what makes it testable — the panel had no component tests, so the rollback would otherwise have shipped unverified. Four cases: happy path (asserting `SetUser` carries `userType: 7` / `userStatus: 1`, which nothing covered before), rollback on failure, the orphan case, a **device-side** `SetCredential` rejection (status `Occupied(3)`) rather than only a transport throw, and a `SetUser` failure that must not attempt a rollback at all. Mutation-tested: deleting the `removeUser` call fails three of them.

I went with rollback rather than inverting the order to `SetCredential`-then-`SetUser`. Inverting is arguably still the better shape — it makes the bad state unrepresentable instead of repaired — but it changes which command creates the user and needs a device to verify against, so it is not something to slip into a review pass. Worth keeping on the list if you ever touch this flow again.

**2. The PIN `.trim()` stays**, per Ingo — trailing whitespace in a PIN is not a real case, and both the create path and `#pinForCommand()` trim, so they remain consistent with each other. That was the actual risk (programming a PIN that could then never unlock), and it does not apply.

**3. The wider log-exposure item is filed as a separate task**, not addressed here: `set_wifi_credentials` (`args.credentials`), `set_thread_dataset` (`args.dataset`, which carries the Thread network key) and the Door Lock Aliro keys are still logged verbatim at debug level, because `redactSensitiveCommandFields` only inspects `args.payload`. Nothing in this PR makes those newly reachable, so it stays out of scope — but the helper's name promises more than it does, and a Thread network key in a pasted support log is worse than a door PIN.

Gates green: format, lint, build, `npm test -w @matter-server/dashboard` (403/403) and `npm test -w @matter-server/ws-client`.

Apollon77:

`#expiringTimeoutSyncedFor` was keyed by node+endpoint, so the expiry field took the value present at mount and then never moved again for as long as the panel stayed on that lock. Another client (or the lock itself) changing `ExpiringUserTimeout` never reached the control, and pressing Save wrote the stale value back over their change.

Replaced the context key with a dirty flag: the field tracks every device report while untouched, stops as soon as the operator types, and resumes after a successful save. The original comment's concern — not clobbering a value someone is mid-edit on — still holds, it is just now keyed on "has been edited" rather than "has been synced once".

453/453 green.

lboue:
Thank you for the corrections you’ve made

Apollon77:
Correct, and subtler than my fix — pushed.

Clearing the dirty flag when `writeExpiringUserTimeout()` resolved raced the attribute cache exactly as described: the `finally` block assigns reactive `_busy`, so `willUpdate()` can run before the `attribute_updated` report lands, find the flag already clear, and put the pre-write cached value back into the field. And if the lock never reports, the field stayed wrong permanently.

The write is now remembered as pending (`#expiringTimeoutAwaitingReport`) and the flag clears only once the cache actually reports that value back. A lock that never reports leaves the entered value standing rather than silently reverting it, which is the safer of the two failure modes. Reset with the rest of the per-context state.

403/403 green.

---

Separately, on the review summary's *"the included fix lacks required backing evidence"*: that is about the PR template, and it is a fair flag now that this PR carries a security fix rather than only a feature. @lboue — the redaction change is a **Fix**, so the template's backing-evidence section applies to it. Reproducing it needs no device: run the server with `--log-level debug`, have a client send `device_command` with a `SetCredential` payload (or a `LockDoor` with `PINCode`), and the base64 PIN appears in the request log. A short log excerpt showing the field before and `[redacted]` after would satisfy it, and is worth having in the PR since the whole point is what does and does not reach the log.

Apollon77:
Agreed on all three — that logging block had grown a fallback and a re-serialization it never needed. Simplified in `cce5cda1`.

**matter.js renders the object itself.** `Logger` formats a plain object as a diagnostic value, bigints included, so there was no reason to produce a string at all:

```
DEBUG WebSocketControllerHandler [c1] WebSocket request { message_id: "7", command: "device_command",
  args: { node_id: 12345678901234567890, endpoint_id: 1, payload: { PINCode: "[redacted]", userIndex: 3 } } }
```

`toBigIntAwareJson` was the wrong tool here twice over: it is the *wire* encoder (with a regex pass to unquote large numbers so the output round-trips), and a log line needs neither property. It stays where it belongs, on outgoing messages.

**The double parse and the fallback both go away by parsing first.** The old shape parsed inside the log closure, then again on the next line, and needed a `<unparseable request, N bytes>` branch for the case where the closure's parse failed. Parsing before the log removes the reason that branch existed — a frame too malformed to redact is now never logged at all, and the parse failure is already reported by the surrounding catch. Net:

```ts
// Parse before logging: an unparseable frame cannot be redacted and may carry a credential.
const request = parseBigIntAwareJson(data) as { message_id: string; command: string; args: any };
logger.debug(`[${connId}] WebSocket request`, redactSensitiveCommandFields(request));
```

**The null checks were load-bearing only because of the parameter type.** `redactSensitiveCommandFields(message: object)` is what forced the caller to prove the parse result was an object before calling it. Taking `unknown` and returning `unknown` removes that from the call site, and both consumers only log the result anyway. Inside, optional chaining plus a nullish default replace the two `=== null || typeof !== "object"` guards:

```ts
const { args } = (message ?? {}) as { args?: { payload?: Record<string, unknown> } };
const payload = args?.payload;
const secrets = Object.keys(payload ?? {}).filter(key => SENSITIVE_PAYLOAD_FIELDS.has(key.toLowerCase()));
if (secrets.length === 0) return message;
```

`Object.keys(payload ?? {})` is safe for every shape a frame can produce — a `null` payload yields no keys, and a string payload yields index keys that cannot match — so no guard is needed to reach it. The field list is a `Set` now rather than an array scan.

I did not move the helper itself onto matter.js `Diagnostic`/`serialize`: it lives in `ws-client`, which is browser-facing and has no `@matter/general` dependency (the dashboard calls the same function from `connection.ts`). It returns a value; the server decides how to render it, which is what lets matter.js do the formatting on that side.

Tests: the repeated `Record<string, any>` casts collapse into one `redactedPayload()` helper, and there is a new case asserting the shapes that used to need the guards (`null`, `undefined`, a string, `{args: null}`, `{args: {payload: null}}`) pass through untouched. Mutation-tested the case-insensitive match — dropping `.toLowerCase()` fails four cases.

Gates green: format, lint, build, `npm test -w @matter-server/ws-client` 99/99 and `npm test -w @matter-server/dashboard` 454/454.

Apollon77:
Right — dropping the thunk moved redaction onto every request. Restored in `3a1063af`.

Verified matter.js defers it rather than assuming:

```
Logger.level = INFO;   l.debug('off', thunk)  ->  calls: 0
Logger.level = DEBUG;  l.debug('on',  thunk)  ->  calls: 1
```

So the value form was costing a scan of every `args.payload` at levels a server actually runs at, for a line that is then thrown away. Final shape:

```ts
// Parse before logging: an unparseable frame cannot be redacted and may carry a credential.
const request = parseBigIntAwareJson(data) as { message_id: string; command: string; args: any };
// Deferred: matter.js calls this only at DEBUG, keeping redaction off the hot path.
logger.debug(`[${connId}] WebSocket request`, () => redactSensitiveCommandFields(request));
```

The comment is there so the arrow function does not read as noise and get inlined away again.

One place this cannot apply: the dashboard's `Connection.sendMessage` calls `console.debug` unconditionally — no level to defer against — so redaction runs on every send there. It returns on the first property lookup when nothing sensitive is present, and the rate is user actions rather than protocol traffic, so it is not worth a gate of its own.

Gates green: format, lint, build, `npm test -w @matter-server/ws-client` 99/99.

Apollon77 (approved):

PR #1035, opened 2026-09-04
feat(dashboard): add EnergyEvse cluster decode + control panel

lboue:
## Summary

Adds a dashboard panel that decodes and controls the **EnergyEvse** cluster (0x99 / 153), following the existing MeterIdentification / CommodityTariff panel pattern.

### Read-only decoding
- State, SupplyState, FaultState (highlighted when active), charging limits, session info
- Feature-gated sections: V2X (bidirectional charging), Charging preferences, SoC reporting, Plug and Charge
- Fixes a pre-existing label bug in the generated cluster/feature descriptions where an embedded acronym (`SoC`, `V2X`) was mis-split by `decamelize` into "So Creporting" / "V2 X"

### Commands
- `Disable`, `EnableCharging`, `StartDiagnostics`, and (when V2X is supported) `EnableDischarging`, each with the relevant inputs (expiry, min/max current)
- A weekly charging schedule editor (`GetTargets` / `SetTargets` / `ClearTargets`) under Charging preferences
- Actions the device's current `SupplyState` would reject (e.g. enabling charging while self-diagnostics are active) are greyed out instead of surfacing a bare `Failure (code 1)`, with a hint pointing at `Disable` when a command fails anyway

## Testing

- `npm run format`, `npm run lint`, `npm run build` — all pass
- `npm test -w @matter-server/dashboard` — 444/444 pass (added `EnergyEvseUtilTest.ts` covering decoding, feature gating, command payload shapes, and the local-datetime ↔ Matter epoch-s conversion)
- Manually exercised against a Matterbridge-simulated EVSE device (`EvseFullFeatures`) with all optional features enabled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This page shows a detailed overview of the changes between versions without the 
 ## **WORK IN PROGRESS**
 
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
+- Enhancement: (lboue) Added a decode and control panel for the EnergyEvse cluster to Dashboard, with a weekly charging schedule editor
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
 
--- a/packages/dashboard/scripts/generate-descriptions.ts
+++ b/packages/dashboard/scripts/generate-descriptions.ts
@@ -23,12 +23,13 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const EVENT_LIST_ATTRIBUTE_ID = 0xfffa;
 
 /**
- * Acronyms with an embedded capital preceded by a lowercase letter (e.g. the "C" in "SoC"). decamelize
- * splits on every lowercase-to-uppercase transition, so it reads that capital as its own word start and
- * lowercases it with the rest: "SoCReporting" -> "so creporting". Protecting the acronym behind a
+ * Names decamelize can't segment as a unit: an acronym with an embedded capital preceded by a lowercase
+ * letter (e.g. the "C" in "SoC"), or letters split apart by a digit (e.g. "V2X"). decamelize splits on
+ * every lowercase-to-uppercase and letter-to-digit transition, so it reads each of those as its own word
+ * start: "SoCReporting" -> "so creporting", "V2X" -> "v2 x". Protecting the acronym behind a
  * digit-suffixed placeholder sidesteps the split, since decamelize does treat digit-to-letter as a boundary.
  */
-const ACRONYMS = ["SoC"];
+const ACRONYMS = ["SoC", "V2X"];
 
 /**
  * Convert camelCase name to human-readable label with a title case.
--- a/packages/dashboard/scripts/generate-descriptions.ts
+++ b/packages/dashboard/scripts/generate-descriptions.ts
@@ -45,7 +45,12 @@ function toLabel(name: string, addSpaces = false): string {
     });
     // Title case: capitalize the first letter of each word
     const label = decamelize(protectedName, " ").replace(/\b\w/g, char => char.toUpperCase());
-    return ACRONYMS.reduce((text, acronym, index) => text.replace(`Acronym${index}`, acronym), label);
+    // Global, and tolerant of a separator decamelize might insert before the digit (e.g. "Acronym 0"):
+    // a plain string .replace() would only fix the first occurrence and only the exact "AcronymN" spelling.
+    return ACRONYMS.reduce(
+        (text, acronym, index) => text.replace(new RegExp(`Acronym\\s*${index}`, "g"), acronym),
+        label,
+    );
 }
 
 interface DeviceType {
--- a/packages/dashboard/scripts/generate-descriptions.ts
+++ b/packages/dashboard/scripts/generate-descriptions.ts
@@ -23,13 +23,28 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
 const EVENT_LIST_ATTRIBUTE_ID = 0xfffa;
 
 /**
+ * Acronyms with an embedded capital preceded by a lowercase letter (e.g. the "C" in "SoC"). decamelize
+ * splits on every lowercase-to-uppercase transition, so it reads that capital as its own word start and
+ * lowercases it with the rest: "SoCReporting" -> "so creporting". Protecting the acronym behind a
+ * digit-suffixed placeholder sidesteps the split, since decamelize does treat digit-to-letter as a boundary.
+ */
+const ACRONYMS = ["SoC"];
+
+/**
  * Convert camelCase name to human-readable label with a title case.
  * e.g., "OnOffLight" -> "On Off Light" when "addSpaces" is set to true, else camelize with first letter uppercase
  */
 function toLabel(name: string, addSpaces = false): string {
-    const words = addSpaces ? decamelize(name, " ") : name;
+    if (!addSpaces) {
+        return name.replace(/\b\w/g, char => char.toUpperCase());
+    }
+    let protectedName = name;
+    ACRONYMS.forEach((acronym, index) => {
+        protectedName = protectedName.split(acronym).join(`Acronym${index}`);
+    });
     // Title case: capitalize the first letter of each word
-    return words.replace(/\b\w/g, char => char.toUpperCase());
+    const label = decamelize(protectedName, " ").replace(/\b\w/g, char => char.toUpperCase());
+    return ACRONYMS.reduce((text, acronym, index) => text.replace(`Acronym${index}`, acronym), label);
 }
 
 interface DeviceType {
--- a/packages/dashboard/src/client/models/descriptions.ts
+++ b/packages/dashboard/src/client/models/descriptions.ts
@@ -8696,7 +8696,7 @@ export const clusters: Record<number, ClusterDescription> = {
             "1": {
                 "bit": 1,
                 "code": "SOC",
-                "label": "So Creporting"
+                "label": "SoC Reporting"
             },
             "2": {
                 "bit": 2,
--- a/packages/dashboard/src/client/models/descriptions.ts
+++ b/packages/dashboard/src/client/models/descriptions.ts
@@ -8711,7 +8711,7 @@ export const clusters: Record<number, ClusterDescription> = {
             "4": {
                 "bit": 4,
                 "code": "V2X",
-                "label": "V2 X"
+                "label": "V2X"
             }
         }
     },
--- a/packages/dashboard/src/pages/cluster-commands/clusters/energy-evse-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/energy-evse-commands.ts
@@ -4,19 +4,101 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { css, html, nothing, type CSSResultGroup } from "lit";
-import { customElement } from "lit/decorators.js";
+import "@material/web/button/filled-button";
+import "@material/web/button/outlined-button";
+import "@material/web/button/text-button";
+import { css, html, nothing, type CSSResultGroup, type PropertyValues, type TemplateResult } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import { showAlertDialog, showPromptDialog } from "../../../components/dialog-box/show-dialog-box.js";
+import { handleAsync } from "../../../util/async-handler.js";
 import { formatDuration } from "../../../util/duration.js";
-import { ENERGY_EVSE_CLUSTER_ID, energyEvseInfo, type SessionInfo } from "../../../util/energy-evse.js";
-import { formatEpochTime } from "../../../util/time.js";
+import {
+    clearChargingTargets,
+    disableEvse,
+    ENERGY_EVSE_CLUSTER_ID,
+    enableCharging,
+    enableDischarging,
+    energyEvseInfo,
+    EVSE_WEEKDAYS,
+    getChargingTargets,
+    setChargingTargets,
+    startDiagnostics,
+    type EditableChargingSchedule,
+    type EditableChargingTarget,
+    type EvseWeekday,
+    type SessionInfo,
+} from "../../../util/energy-evse.js";
+import { errorText } from "../../../util/error-text.js";
+import { formatEpochTime, fromLocalDateTimeInputValue } from "../../../util/time.js";
 import { BaseClusterCommands } from "../base-cluster-commands.js";
 import { registerClusterCommands } from "../registry.js";
 
+const MAX_SCHEDULES = 7;
+const MAX_TARGETS_PER_SCHEDULE = 10;
+
+function minutesToTimeInputValue(minutes: number): string {
+    const hours = Math.floor(minutes / 60) % 24;
+    const mins = minutes % 60;
+    return `${String(hours).padStart(2, "0")}:${String(mins).padStart(2, "0")}`;
+}
+
+function timeInputValueToMinutes(value: string): number | undefined {
+    const match = /^(\d{2}):(\d{2})$/.exec(value);
+    if (match === null) return undefined;
+    const hours = Number(match[1]);
+    const minutes = Number(match[2]);
+    return hours * 60 + minutes;
+}
+
 /**
- * Read-only decoding panel for the EnergyEvse cluster (ID: 0x99 / 153).
+ * Decoding and command panel for the EnergyEvse cluster (ID: 0x99 / 153).
  */
 @customElement("energy-evse-cluster-commands")
 export class EnergyEvseClusterCommands extends BaseClusterCommands {
+    @state() private _busy = false;
+    @state() private _formError?: string;
+
+    @state() private _chargeNoExpiry = true;
+    @state() private _chargeUntil = "";
+    @state() private _minChargeCurrentA = 6;
+    @state() private _maxChargeCurrentA = 16;
+
+    @state() private _dischargeNoExpiry = true;
+    @state() private _dischargeUntil = "";
+    @state() private _maxDischargeCurrentA = 16;
+
+    @state() private _schedules?: EditableChargingSchedule[];
+    @state() private _scheduleBusy = false;
+    @state() private _scheduleError?: string;
+
+    #context?: string;
+    /** Bumped only on a node/endpoint context change, so a command run against the old context can't clobber the new one's UI state. */
+    #busyGeneration = 0;
+
+    override willUpdate(changedProperties: PropertyValues) {
+        super.willUpdate(changedProperties);
+        if (!this.node || this.cluster !== ENERGY_EVSE_CLUSTER_ID) return;
+
+        const context = `${String(this.node.node_id)}/${this.endpoint}`;
+        if (this.#context === context) return;
+        this.#context = context;
+        this.#busyGeneration++;
+        this._busy = false;
+        this._formError = undefined;
+        this._chargeNoExpiry = true;
+        this._chargeUntil = "";
+        this._dischargeNoExpiry = true;
+        this._dischargeUntil = "";
+        this._schedules = undefined;
+        this._scheduleBusy = false;
+        this._scheduleError = undefined;
+
+        const info = energyEvseInfo(this.node.attributes, this.endpoint);
+        this._minChargeCurrentA = info.minimumChargeCurrentA ?? this._minChargeCurrentA;
+        this._maxChargeCurrentA = info.maximumChargeCurrentA ?? this._maxChargeCurrentA;
+        this._maxDischargeCurrentA = info.maximumDischargeCurrentA ?? this._maxDischargeCurrentA;
+    }
+
     override render() {
         if (!this.node || this.cluster !== ENERGY_EVSE_CLUSTER_ID) return nothing;
         const info = energyEvseInfo(this.node.attributes, this.endpoint);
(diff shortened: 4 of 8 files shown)


Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`). All eight open Copilot comments were correct and are addressed. Copilot missed the blocker, though:

**`dayOfWeekForSequence` never decoded.** `decodeChargingTargetSchedule` read it with `asObject(...)` + `pickBoolean(...)`, but the server collapses every bitmap to an integer on the way out (`Converters.ts`, `ConvKind.Bitmap`). I ran the real converter against a synthetic `GetTargetsResponse` to confirm: a Mon+Wed schedule comes back as `dayOfWeekForSequence: 10`, not `{monday: true, wednesday: true}`. So "Load current" returned every stored schedule with **no** day selected, and the panel's own "Select at least one day for every schedule" validation then refused to save it — the load→save round trip was impossible.

The write direction happened to work (matter.js's bitmap schema accepts the object form), which is exactly why the self-consistent tests missed it: `EnergyEvseUtilTest.ts` asserted the object shape as *expected output*, a shape the server cannot produce. Both directions now use the integer, bit 0 Sunday through bit 6 Saturday, matching `thermostat-schedule.ts` and `door-lock.ts`, which already decode the same weekday bitmap that way. I verified with the converter that a number is accepted inbound too. Mutation-tested: shifting the bit index by one fails three tests.

The rest of the commit, grouped the way the findings cluster:

**Form state had no empty value.** `Number("")` is `0` and `Number.isFinite(0)` is true, so clearing a target's SoC field set `targetSoC = 0` — telling the EVSE to charge to 0% — and Lit wrote `"0"` straight back, so the field could not be cleared at all. Same in `#parsePositiveNumber` for the current limits. One `#optionalNumber()` now maps `""` to `undefined`.

**SOC as an exclusive editor switch.** Spec §9.3.7.6.2 says AddedEnergy takes precedence when the EVSE "does not support the SOC feature **or cannot obtain the SoC of the vehicle**" — so hiding the kWh input on a SoC-capable device removes the fallback the spec designates for that exact device. Both inputs are offered now; setting either goal clears the other.

**Validation stopped short of what the device enforces.** Added the two the spec constrains and the panel did not: a weekday may appear in only one schedule (§9.3.9.5.1), and `ChargingTargetStruct`'s `O.a+` group means a target must carry a SoC or an added-energy goal. Both previously produced a bare `Failure` from the device.

**`renderWhenOffline` copied from decode-only panels.** `meter-identification` and `commodity-tariff` have no commands, so the template you followed had no offline gate and no context-guarded dialog helper. Every command button now also checks `node.available` (as `access-control`, `binding` and `icd-management` do), and `#reportFailure` takes the busy generation so an old device's failure cannot pop over the new one — `cluster-commands/AGENTS.md` is explicit that the guard goes before the dialog.

Smaller items: `fromLocalDateTimeInputValue` rejects dates outside `0..MATTER_EPOCH_MAX_SECONDS` and the inputs carry `min`/`max` (a 1999 date became a negative uint32 that failed at serialization rather than in the form — `door-lock.ts:dateRangeError` already had this pattern); `StartDiagnostics` is optional and is now gated on `AcceptedCommandList` like `door-lock.ts:supportsCommand` does for `UnlockWithTimeout`; accessible names on the two `datetime-local` inputs and the target `time` input; `.schedule-card` moved to `surface-container-highest` (`-high` is `#fff` in the light theme, the same white as the panel behind it, so the card boundary vanished); and a CHANGELOG entry, which the branch had none of.

Two things I checked and left alone:

- **The decamelize acronym fix has no blast radius.** `toLabel(name, false)` is byte-identical to the old behaviour, and grepping every `name:`/`title:` in `@matter/model` yields only `NextChargeTargetSoC`, `TargetSoC` and `V2X`. Attribute labels don't go through `toLabel` at all. Note the `\s*` in the restore regex is dead (decamelize never separates before a digit) and the adjacent `Rfid` feature label is broken by the same underlying limitation — not worth widening here.
- **`canStartDiagnostics` is only half-founded.** Gating StartDiagnostics on `SupplyState == Disabled` is confirmed by the reference implementation (`EnergyEvseDelegateImpl.cpp:361`). The mirror rule — greying out EnableCharging/EnableDischarging while `DisabledDiagnostics` — is not: neither the spec text nor `EnergyEvseCluster::HandleEnableCharging` rejects it. That control is disabled on a guess. Harmless (the hint tells the user to press Disable), but you may want to drop it.

Still open, for your call: `_handleSaveSchedule` and `_handleLoadSchedule` leave the schedule editors interactive while in flight, so a `GetTargets` response overwrites edits made while it was loading. Reshaping that means disabling the editor during the operation or reconciling the response — I did not want to change the editing model without you.

Apollon77:
## Proposal: the schedule editor's in-flight edit race

Follow-up to the review above — this is the one item I did not touch, because fixing it changes how the editor behaves and that is a call for you, not me.

### What happens today

`_schedules` is a single `@state` array that is both **the editor's contents** and **the last thing the device told us**. `_scheduleBusy` disables the three action buttons (Load current / Clear all / Save schedule) but nothing inside a schedule card: the day chips, the time inputs, the SoC/kWh inputs, Add target, Remove target and Remove schedule all stay live while a command is in flight.

Two ways that goes wrong:

1. **Load overwrites edits.** Press *Load current*, edit a target while `GetTargets` is in flight, and `this._schedules = schedules` (`_handleLoadSchedule`) discards the edit with no indication. The window is a full round trip to the device.
2. **Save silently under-sends.** `_handleSaveSchedule` reads `this._schedules` once, before the `await`. Edits made while `SetTargets` is in flight stay on screen but were never in the payload — so the editor afterwards shows a schedule the EVSE does not have, and nothing says so.

Neither corrupts the device. Both leave the user believing something is stored that is not, which for a charging schedule means the car does not charge when they expect.

### Three options

**A — Disable the editor while busy.** Thread `_scheduleBusy` into `_renderScheduleEditor`/`_renderTargetEditor` and disable every input and button inside the cards, the way the action row already is. ~10 lines, no new state, no behaviour to explain. Cost: the editor freezes for the duration of a round trip, which on a slow lock/EVSE is noticeable, and a user mid-keystroke loses focus.

**B — Guard the write, leave the editor live.** Keep editing enabled; on the `GetTargets` response, only assign `_schedules` if the user has not touched it since the request started (a dirty flag or an edit counter captured before the `await`, checked after — the same shape as the existing `#busyGeneration`). If they have, keep their edits and surface `"The device's stored schedule was loaded but your unsaved edits were kept — press Load current again to discard them."` For save, re-read `_schedules` after the await and compare against what was sent; if they differ, say `"Edited while saving — press Save again to store the current version."` Cost: a little more state and two messages to word, but the editor never freezes.

**C — Separate the two roles.** `_deviceSchedules` (last known device state) and `_draftSchedules` (what the user is editing), with an explicit "discard changes / reload from device" action and a visible dirty marker. This is the shape that makes both bugs unrepresentable rather than guarded, and it is what the panel wants if schedule editing grows (per-schedule save, diff against device, etc.). Cost: the largest diff of the three and a real UI change — arguably its own PR.

### Recommendation

**A for this PR, C when the editor next grows.** A is small enough to not risk the rest of the PR, it matches what the action row already does so there is nothing new for a user to learn, and the freeze is bounded by one round trip. B trades a visible freeze for two failure messages that are easy to get subtly wrong, which is a worse deal at this size. C is right but it is a different change.

If you would rather keep the editor live, B is fine too — I would just want the "kept your edits" path covered by a test, since it is the branch nobody exercises by hand.

Happy to implement whichever you pick, here or as a follow-up.

Apollon77:
One more push to this branch, closing the last item from my review that was not a proposal.

**Dropped the `DisabledDiagnostics` gate on Enable Charging / Enable Discharging.** I had flagged this as disabled-on-a-guess; Ingo agreed, so it is gone. To be explicit about the evidence, since the two halves of that rule look symmetric but are not:

- **`StartDiagnostics` only while `SupplyState == Disabled` — kept.** Confirmed in the reference implementation (`connectedhomeip`, `EnergyEvseDelegateImpl.cpp:361`), which rejects it otherwise.
- **`EnableCharging`/`EnableDischarging` blocked while `DisabledDiagnostics` — removed.** Nothing enforces this: not the cluster's spec text, not `EnergyEvseCluster::HandleEnableCharging` (`src/app/clusters/energy-evse-server/EnergyEvseCluster.cpp:472-487`), not the delegate. A device that accepts the command during self-diagnostics could not be given it.

The failure hint that points at *Disable* stays on all three commands — it costs nothing on a device that does reject, and Matter only hands the controller a bare status code, so the hint is the only guidance available when one does. `diagnosticsActive` had no other consumer, so it is out of `EnergyEvseInfo` too.

Everything else from the review is in the earlier commits; the schedule in-flight edit race is written up as a separate proposal comment above for you to pick from.

lboue:
Going with **option A**. I pushed in 97642df. `_scheduleBusy` now disables every input/button inside the schedule cards (days, time, SoC/kWh, Add/Remove target, Remove schedule), same gate the action row already had, plus `:disabled` styling since those elements set their own colors/cursor.

Apollon77:
Five of Copilot's six new comments are correct and are fixed; one I am pushing back on with evidence.

**First, an apology and a correction:** my push briefly force-overwrote @lboue's `97642df2` ("disable EnergyEvse schedule editor while a request is in flight" — option A from the proposal). I noticed immediately and restored it by cherry-pick, so it is back on the branch as `0f9b5ed1` with its original authorship, merged with my constant rename. Nothing was lost. Cause: I ran `git fetch` immediately before `git push --force-with-lease`, which refreshes the lease and defeats the protection it exists to provide. My mistake, not a tooling one.

### Three findings, one cause — reshaped rather than patched

`:763` (empty target list), `:879` (SoC vs added energy) and `:620` (current input) all landed on the same shape in the round *after* I had already patched that shape once, so I stopped adding guards and moved the rules instead.

The cause: validation was a list of ad-hoc checks inside `_handleSaveSchedule` that did not know the device's feature map, so every rule had to be hand-written and two of the ones present were simply wrong about the conformance. I verified both against `@matter/model`:

```js
Field({ name: "TargetSoC",   id: 1, type: "percent",    conformance: "SOC, O.a+" }),
Field({ name: "AddedEnergy", id: 2, type: "energy-mWh", conformance: "[SOC], O.a+" }),
Field({ name: "ChargingTargets", id: 1, type: "list", conformance: "M", constraint: "max 10" }),
```

- **`ChargingTargets` has no minimum.** An empty target list is legal and is how a day's targets are cleared. Rejecting it made a loaded response unsavable. Copilot correct.
- **`TargetSoC` is `SOC, O.a+` — mandatory once the SOC feature is active**, and `AddedEnergy` is `[SOC], O.a+`, so it *accompanies* a SoC goal rather than replacing one. My earlier "setting either clears the other" was wrong for SoC devices: entering energy cleared the SoC the device requires, and the save check accepted it, so `SetTargets` answered `InvalidCommand`. Copilot correct, and it caught a defect I introduced.

`chargingScheduleError(schedules, soCSupported)` in the util now owns every rule — day selection, day uniqueness across schedules, the list caps, the time range, and the per-feature goal requirement — which is what lets the conformance decide instead of the editor. The `max 7` / `max 10` caps moved there too, so the button limits and the payload check read the same constants. Ten test cases, mutation-tested: relaxing the SoC requirement and re-adding the empty-list rejection each fail a test.

**`:620` current input — correct, fixed.** The three current fields held a parsed number while the DOM held whatever was typed, so an invalid entry left the two disagreeing and Enable sent the previous value while the box showed the new one. They hold raw text now and are validated on submit, like the target fields.

### Two more, both correct

- **`energy-evse.ts:193`** — `canStartDiagnostics` treated an unread `SupplyState` as permission to run. An unread mandatory attribute is not evidence the command would be accepted; it is `false` until the state is known.
- **`time.ts:47`** — `new Date("2027-03-14T02:30")` in a zone with a spring-forward gap silently becomes `03:30`, so the command would carry an expiry an hour off what was typed. The parsed instant is now round-tripped back through the local formatter and rejected if it changed. The suppressed `time.ts:51` comment was right too: the `min`/`max` bounds were UTC clock values on inputs that operate in local time, so they blocked valid instants at one end and permitted invalid ones at the other — both are derived with the local formatter now.

### One I am not taking: `:347` / `:413` EnableCharging during DisabledDiagnostics

This is the gate I *removed* two commits ago, deliberately, and I do not think it should come back. The evidence:

- **Spec text** for `EnableCharging` places no constraint on `SupplyState`.
- **`EnergyEvseCluster::HandleEnableCharging`** (`connectedhomeip`, `src/app/clusters/energy-evse-server/EnergyEvseCluster.cpp:472-487`) does not check it.
- **`EnergyEvseDelegateImpl`** does not reject it either — whereas the *same* delegate at `EnergyEvseDelegateImpl.cpp:361` **does** enforce the `Disabled` precondition for `StartDiagnostics`, which is why that gate stays.

So the two halves look symmetric but only one is founded. Greying out a command that the dominant implementation accepts costs a user a working action to prevent a failure that does not occur. Copilot also notes this "contradicts the PR's stated state-based greying behavior" — that part is fair, but the fix is to the description, not the code: @lboue, the *Commands* bullet should drop the claim that enable actions grey out during self-diagnostics and keep only the `StartDiagnostics` case.

The failure hint pointing at *Disable* stays on all three commands regardless — Matter hands the controller only a bare status code, so on a device that *does* reject, that hint is the only guidance available.

Happy to be overruled if someone has a device that actually rejects it — that would be a capture worth having.

Gates: format, lint, build, `npm test -w @matter-server/dashboard` 481/481.

lboue:
Re `:347`/`:413` — I don't think this one is symmetric the way it looks. `EnergyEvseDelegateImpl` does reject it, one level down from where the delegate's own `EnableCharging()`/`EnableDischarging()` bodies would show it:

```
EnergyEvseDelegate::EnableCharging()                    EnergyEvseDelegateImpl.cpp:116
  -> HandleStateMachineEvent(ChargingEnabledEvent)                          :845/867
  -> HandleChargingEnabledEvent()                                           :1010
  -> CheckFaultOrDiagnostic()                                               :1013
       -> Status::Failure when SupplyState == kDisabledDiagnostics     :1002-1006
```

`HandleDischargingEnabledEvent()` calls the same `CheckFaultOrDiagnostic()` guard for `EnableDischarging`. Neither top-level function checks `SupplyState` itself — they just validate current limits and hand off to the state machine — which is why reading `EnableCharging()`/`EnableDischarging()` alone looks unguarded. The guard is in the state-machine event handler underneath, not the command entry point.

Also checked it isn't a recency thing: `CheckFaultOrDiagnostic()` predates the EVSE code-driven conversion. It's already in `examples/energy-management-app` back at `4abbe28` ("EVSE add Session and Event support"), so this isn't new behavior the earlier read could have missed by version.

Restored `diagnosticsActive` and the two `?disabled` bindings, rebased on `0f9b5ed` . I pushed as 4c930c2. format/lint/build/test all green. If you've got a device that actually accepts the command during diagnostics, that'd be worth a test-event-trigger capture, and I'll revert without a fight.

Apollon77:
@lboue you are right and I was wrong — thanks for chasing it one level further instead of just accepting my pushback.

I verified your trace in the local checkout rather than take either side on assertion:

```c
// examples/evse-app/evse-common/src/EnergyEvseDelegateImpl.cpp:994
Status EnergyEvseDelegate::CheckFaultOrDiagnostic()
{
    if (GetFaultState() != FaultStateEnum::kNoError)          { ... return Status::Failure; }
    if (GetSupplyState() == SupplyStateEnum::kDisabledDiagnostics) { ... return Status::Failure; }
    return Status::Success;
}
```

called from `HandleChargingEnabledEvent()` (:1013) and `HandleDischargingEnabledEvent()` (:1075). My check stopped at `EnergyEvseCluster::HandleEnableCharging` and the delegate's `EnableCharging()` entry point and concluded "the delegate does not reject it either" — that was an inference from an incomplete read that I then reported as verified. Exactly the failure mode my own report warns about; I got the evidence tier wrong, not just the answer. Your gate is restored and stays.

**Reading it properly costs two more findings**, both from the same guard, so I have pushed one follow-up commit (`bf0ee0d7`) on top of yours rather than touching your commit:

**1. `Disable` is refused during diagnostics too.** `HandleDisabledEvent()` (:1132) calls the same `CheckFaultOrDiagnostic()`. So the hint — *"click Disable first"* — told the operator to press a button that cannot work in that state. Nothing client-side leaves `DisabledDiagnostics`: only the device does, via `HwDiagnosticsComplete()` (:818), which puts it back to `Disabled`. Both the button title and `DIAGNOSTICS_OR_ALREADY_ENABLED_HINT` said otherwise.

**2. A fault blocks the same three commands, and nothing checked it.** `CheckFaultOrDiagnostic()` fails on any `FaultState != kNoError`. The panel highlighted the fault but still offered Disable, Enable Charging and Enable Discharging — on a faulted EVSE, all three were clickable and all three failed with a bare status code.

Since it is one predicate in the reference implementation, it is now one predicate here: `supplyCommandsBlockedReason` on `EnergyEvseInfo` returns why the three commands would be refused, or `undefined`, and gates all three with the reason as the control's `title`. A fault outranks diagnostics, because that is the condition the operator has to act on first. Test covers all four combinations; mutation-tested by inverting the fault branch.

`StartDiagnostics` keeps its own separate `SupplyState == Disabled` gate — that one is enforced in `StartDiagnostics()` (:356) rather than by the shared guard.

Also worth noting for the description: with this in, the *Commands* bullet's claim about state-based greying is accurate again, so ignore the correction I asked for in my previous comment.

Gates: format, lint, build, `npm test -w @matter-server/dashboard` 420/420 on a clean build.

Apollon77:
Six of the seven are correct and are fixed; the seventh is a description item I have left open for @lboue.

**Two real defects, both the same class as the ones from the last round:**

- **`:871` cleared time field.** Clearing the time input made the parser return `undefined` and the handler simply returned, leaving the previous `timeMinutes` in `_schedules`. The control read blank while Save transmitted the old time. `timeMinutes` is now `number | undefined`, an empty field genuinely unsets it, and `chargingScheduleError` rejects it. `setChargingTargets` throws on an unset time rather than defaulting, since validation gates every send and reaching it would be a bug.
- **`util:372` fractional SoC.** `TargetSoC` is a `percent`, a whole-number uint8, but the control accepted `50.5` and the check only asked whether the value was defined — so `SetTargets` failed during encoding rather than in the form. Validated as an integer, and the input carries `step="1"`.

**Three stale comments and a wrong noun, all mine:**

- `util:48` and `:100` still said self-diagnostics is cleared by `Disable`. It is not — that was the whole point of the correction two commits ago, and I updated the behaviour without updating the prose describing it. They now say the device leaves the state by finishing.
- `util:386` still called `targetSoC` and `addedEnergyKWh` mutually exclusive, contradicting both the conformance logic and its own test. They now describe the feature-dependent rule.
- `util:348` reported *"A lock stores at most 7 schedules"* in an EVSE panel — copy-paste from `door-lock`.

**The DST test asserted nothing, which is the finding I am most glad of.** In a zone without a spring-forward gap `parsed` is defined and only a trivial round-trip was checked; in CI (UTC) the guard could have been deleted with the suite still green. It now runs the conversion in a child process pinned to `America/New_York` and asserts that the non-existent `2027-03-14T02:30` is rejected while `01:30` and `03:30` are not. Reverting the round-trip guard fails it — I checked.

**Left open: `:52`, the hint vs. the PR description.** The hint text is now correct — the panel disables `Disable` during faults and diagnostics because the reference delegate refuses it there, and neither state can be ended from the dashboard. It is the *description* that still promises a `Disable`-based recovery. @lboue, the Commands bullet should say that actions the EVSE would refuse are greyed out with the reason, and drop the "hint pointing at `Disable`" wording. I have not resolved that thread since the change is yours.

Gates: format, lint, build, `npm test -w @matter-server/dashboard` 422/422 on a clean build.

Apollon77:

**Correction — the greying-out behaviour, and how a blocked EVSE recovers.** The Commands section says actions the current `SupplyState` would reject are greyed out "with a hint pointing at `Disable` when a command fails anyway". Two parts of that are now wrong:

- The set of blocked states is wider. `Disable`, `EnableCharging` and `EnableDischarging` are all guarded by the reference delegate's single `CheckFaultOrDiagnostic()`, which refuses them **while any fault is active** as well as during self-diagnostics. The panel previously only highlighted a fault while still offering all three.
- `Disable` is **not** the recovery. It is refused in those states too (`HandleDisabledEvent()` calls the same guard). Self-diagnostics ends only when the device finishes it (`HwDiagnosticsComplete()`), and a fault has to clear on the device; neither can be ended from the dashboard.

So: the three supply commands grey out together with the reason shown as the control's title, and the failure hint says the condition has to clear on the device rather than pointing at a button that cannot work. `StartDiagnostics` keeps its own separate gate — only while `SupplyState == Disabled`, which the delegate does enforce in `StartDiagnostics()`.

**Behaviours added during review, all worth knowing when reading the diff:**

- **`dayOfWeekForSequence` is an integer bitmap**, not an object of booleans — the server collapses every bitmap on the way out. Decoding it as an object meant "Load current" returned every stored schedule with no day selected, and the panel's own validation then refused to save it, so the load→save round trip was impossible. Both directions use the integer now (bit 0 Sunday … bit 6 Saturday), matching `thermostat-schedule.ts` and `door-lock.ts`.
- **Schedule validity is derived from the cluster conformance**, in `chargingScheduleError()` in the util rather than ad-hoc checks in the save handler: `TargetSoC` is `SOC, O.a+` (mandatory once the SOC feature is active) and `AddedEnergy` is `[SOC], O.a+` (accompanies it, does not replace it); `ChargingTargets` is `max 10` with **no minimum**, so an empty target list is legal and is how a day is cleared; a weekday may appear in only one schedule; a target SoC must be a whole percent.
- **Every numeric field keeps what was typed** and is validated on submit. Previously the parsed state and the DOM could disagree, so an invalid entry left the box showing one value while Enable sent another, and clearing a SoC field wrote `0` — telling the EVSE to charge to 0 %.
- **`renderWhenOffline` gating.** The panel stays mounted for its cached decoding, so every command is now also gated on `node.available`, and the shared failure dialog is context-guarded.
- **Date handling.** `datetime-local` values outside `0..MATTER_EPOCH_MAX_SECONDS` are rejected in the form rather than at serialization, the bounds are derived in the viewer's zone, and a local time inside a DST spring-forward gap is refused instead of being silently moved an hour.
- **`StartDiagnostics` is optional** and is gated on `AcceptedCommandList`, and an unread `SupplyState` no longer counts as permission to run it.
- **The schedule editor is disabled while a request is in flight** (@lboue's commit), so a `GetTargets` response cannot overwrite an in-progress edit and a `SetTargets` cannot silently under-send one.
- **Adds a CHANGELOG entry**, which the branch originally had none of.

Apollon77 (approved):

PR #1050, opened 2026-09-09
feat(dashboard): decode the DeviceEnergyManagement Forecast attribute

lboue:
## Type of change

## Description

Adds a read-only Dashboard panel decoding the DeviceEnergyManagement cluster's (0x98) `Forecast` attribute — the piece that publishes what an energy smart appliance is about to draw or generate, slot by slot.

Each slot in `ForecastStruct.Slots` resolves to a row with its status (completed / active / scheduled, relative to `ActiveSlotNumber`), power, duration and energy:

- Slots carry durations, not instants, so clock times are accumulated from `Forecast.StartTime` across preceding slots. A slot with no usable duration stops the chain instead of guessing the ones after it.
- The active slot is timed by its own `ElapsedSlotTime`/`RemainingSlotTime` (the only figure that tracks a device drifting from its plan); every other slot falls back to `DefaultDuration`.
- Energy uses `NominalEnergy` when the device reports it, otherwise derives it from `NominalPower × duration` and flags the result as estimated.
- Negative power is generation (PV, batteries discharging): totals split into `consumedEnergyWh` / `generatedEnergyWh` rather than netting them together.
- `StartTimeAdjustment`'s earliest/latest bounds and each slot's power/duration adjustment limits are surfaced for shiftable loads (EVSE charging windows, etc.).
- Slot `CostStruct` entries reuse CommodityTariff's currency-scaling logic (now exposed as `currencyInfo()`) to render a priced amount.

Covers the three use cases this was built for: 
 * a dishwasher's SFR-based state forecast, 
 * a solar PV device's PFR-based generation forecast,
 * an EVSE's shiftable charging session with adjustment limits.

Known limitation: 
* `ManufacturerESAState` (used under SFR) is an opaque per-vendor integer
* The panel shows it as a chip (`state N`) rather than a name, since naming it would need a vendor/product-specific mapping.

## Backing evidence

N/A — this is a feature addition, not a fix; no defect is being corrected.

- Related issue: #

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This page shows a detailed overview of the changes between versions without the 
 
 ## **WORK IN PROGRESS**
 
+- Enhancement: (lboue) Added a Forecast panel for the DeviceEnergyManagement cluster to Dashboard, decoding per-slot power/energy, timing and adjustment limits
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
--- a/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-commands.ts
@@ -179,12 +179,14 @@ export class DeviceEnergyManagementClusterCommands extends BaseClusterCommands {
 
     /** Nominal draw when the device commits to one, otherwise the band it stays inside. */
     private _slotPower(slot: ForecastSlotInfo, info: DeviceEnergyManagementInfo): TemplateResult | typeof nothing {
-        const generating = (slot.nominalPowerW ?? slot.energyWh ?? 0) < 0;
+        const generating = (slot.nominalPowerW ?? slot.energyWh ?? slot.minPowerW ?? slot.maxPowerW ?? 0) < 0;
         const value =
             slot.nominalPowerW !== undefined
                 ? formatPower(Math.abs(slot.nominalPowerW))
                 : slot.minPowerW !== undefined && slot.maxPowerW !== undefined
-                  ? `${formatPower(Math.abs(slot.minPowerW))}–${formatPower(Math.abs(slot.maxPowerW))}`
+                  ? generating
+                      ? `${formatPower(Math.abs(slot.maxPowerW))}–${formatPower(Math.abs(slot.minPowerW))}`
+                      : `${formatPower(Math.abs(slot.minPowerW))}–${formatPower(Math.abs(slot.maxPowerW))}`
                   : undefined;
         if (value === undefined) return nothing;
         return html`
--- a/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-commands.ts
@@ -145,6 +145,11 @@ export class DeviceEnergyManagementClusterCommands extends BaseClusterCommands {
                     </div>
                     <div class="slot-sub">${this._slotDetail(slot)}</div>
                     ${
+                        this._slotAdjustment(slot) !== undefined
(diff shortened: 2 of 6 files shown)


Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`). The decoding itself is solid — every `ForecastStruct` tag (0-8), `SlotStruct` tag (0-17) and `CostStruct` tag matches `@matter/model` exactly, including the two the tag-keyed wire format makes easy to slip (`Costs` = 13, `MinPowerAdjustment` = 14); the mW/mWh scaling, the epoch-s offset and the bigint handling are all right; and the `currencyInfo()` extraction is behaviour-identical for CommodityTariff, its only other caller.

Both Copilot comments are correct.

**The adjustment limits one is the more serious**, and it belongs to a pattern rather than a missing row: the util decodes 18 slot fields and 7 feature flags, and the panel consumed roughly half of them. `minPowerAdjustmentW`, `maxPowerAdjustmentW`, `minDurationAdjustmentSeconds` and `maxDurationAdjustmentSeconds` were decoded, typed *and* unit-tested, but no render path read them — so the EVSE shiftable-load case the PR names as covered showed none of its headline data, and both the PR body ("each slot's power/duration adjustment limits are surfaced") and the CHANGELOG ("and adjustment limits") were false against the code. There is now an `Adjustable: power … · duration …` sub-line per slot, which also retires the four unit-test assertions that were certifying an invisible feature.

**The estimated flag** was forecast-wide (`slots.some(...)`), so on a bidirectional battery forecast one estimated consuming slot made a fully-reported generation total read "(est.)". Tracked per direction in the util now.

Also in the commit:

- **`ESA_TYPE_NAMES` had a fabricated entry.** `ESATypeEnum` is 0–13 plus 255; there is no 14. A device reporting an unrecognised type 14 rendered as a confident "Water pump" instead of falling through to `Unknown (14)`. Every other id matches.
- **The running slot had no textual state.** `_slotDetail` pushed `STATUS_LABELS[slot.status]` only in the `else` branch, and the `●` marker is `aria-hidden="true"` — so a screen reader heard "Slot 3 · Scheduled · 20 min" for every slot except the one whose state matters most. The label is unconditional now.
- **The chips were invisible in the light theme.** `.chip` used `--md-sys-color-surface-container-high`, which is `#fff` in light mode, on a panel that is also `#fff`, with no border. So the `state N` chip — the PR's stated presentation for `ManufacturerESAState` — and the `pausable` chip rendered as unstyled text. Moved to `secondary-container`, which differs in both themes.
- **The active-slot clock test proved nothing.** In every fixture, elapsed + remaining happened to equal `DefaultDuration` (dishwasher 1320+1080 = 2400; solar 600+3000 = 3600; the EVSE fixture has a null `ActiveSlotNumber`), so deleting the whole `if (status === "active" …)` line left the suite green — including the test named *"times the running slot by its own clock and the others by their default duration"*. There is now a drifting-device fixture where the two differ, and it asserts the downstream `startTime` shift too. I mutation-tested it along with the per-direction estimated flags: three tests fail, restored.

Two things I looked at and deliberately did not guard:

- `_slotPower` uses `Math.abs` on a min/max band, so a BESS slot spanning zero (`MinPower = -3000`, `MaxPower = +2000`) with no `NominalPower` would render the charge half as generation. But `NominalPower` is conformance `PFR`, same as MinPower/MaxPower, so on a conformant device the min/max-only branch cannot occur — theoretical, and a dedicated guard would just be surface for the next review round.
- Same for the PR's "a slot with no usable duration stops the chain" claim: `MinDuration`, `MaxDuration`, `DefaultDuration`, `ElapsedSlotTime` and `RemainingSlotTime` are all `M` on `SlotStruct`, so no conformant device produces the input the guard exists for. The code is right and cheap, the claim is just hollow in practice.

One nit left for you: `decodeCost` renders `CostStruct.Currency` for any `CostType`, so a device setting Currency alongside `CostType = GhgEmissions` yields "GHG emissions 0.1579 €".

Apollon77:
Correct — fixed and pushed. Good catch, and it is the sharper version of something I had already looked at without seeing the consequence.

`slotSeconds` counted a missing duration as zero, so for the `[600, unknown, 600]` case the timing chain correctly stopped at slot 1 while `durationSeconds` still reported 1200 s — a partial sum presented as the whole forecast's length, which is worse than showing nothing because it looks authoritative. The fallback now applies only when every slot contributes, and the existing unknown-duration test asserts `durationSeconds === undefined` rather than just checking the chain. Mutation-tested: relaxing `every` to `some` fails it.

406/406 on a clean build.

Apollon77 (approved):

PR #1055, opened 2026-09-12
feat(dashboard): add DeviceEnergyManagementMode command panel

lboue:
## Type of change

## Description

Adds a Dashboard command panel for the `DeviceEnergyManagementMode` cluster (0x9F) — a Mode Base derivative (`SupportedModes`/`CurrentMode`/`StartUpMode`/`OnMode` + `ChangeToMode`) that lets an energy management appliance be steered between optimization strategies (no optimization / device-local / grid-driven, etc.).

- Decodes `SupportedModes` (`ModeOptionStruct` list) into label + mode number + `ModeTag` names, covering both the common Mode Base tags (Auto…Day) and the cluster-specific ones (`NoOptimization`, `DeviceOptimization`, `LocalOptimization`, `GridOptimization`), confirmed against the cluster definition in `@matter/model` (`device-energy-management-mode.element.ts` / `mode-base.element.ts`).
- Shows the current mode (resolved to its label), and `StartUpMode`/`OnMode` when the device exposes them.
- The panel sends `ChangeToMode` via `client.deviceCommand()` directly (instead of the base `sendCommand()` helper, which discards the response) so the `ChangeToModeResponse`'s `Status`/`StatusText` is decoded and shown — success or the device's rejection reason.
- Unlike the read-only Forecast panel from #1050, this panel issues an active command, so it isn't registered with `renderWhenOffline`.

Companion to #1050 (DeviceEnergyManagement Forecast panel) but opened separately since that PR is already ready for review.

## Backing evidence

N/A — this is a feature addition, not a fix; no defect is being corrected.

- Related issue: #

## Checklist

- [x] `npm test` passes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ This page shows a detailed overview of the changes between versions without the 
 
 - Enhancement: (lboue) Added a command panel for the DoorLock cluster to Dashboard
 - Enhancement: (lboue) Added Presets and Thermostat Suggestions (Thermostat cluster PRES/TSUGGEST features) panels to Dashboard
+- Enhancement: (lboue) Added a command panel for the DeviceEnergyManagementMode cluster to Dashboard
 - Fix: BLE proxy connections are pinged every 15 seconds and terminated after 45 to 60 seconds of silence, so a proxy client that loses power is detected instead of staying registered indefinitely
 
 ## 1.4.0 (2026-08-07)
--- a/packages/dashboard/public/index.html
+++ b/packages/dashboard/public/index.html
@@ -44,6 +44,8 @@
     --md-sys-color-on-error-container: #410e0b;
     --md-sys-color-tertiary: #7d5260;
     --md-sys-color-on-tertiary: #fff;
+    --md-sys-color-tertiary-container: #ffd8e4;
+    --md-sys-color-on-tertiary-container: #31111d;
     --md-sys-color-primary-container: #d0e4ff;
     --md-sys-color-on-primary-container: #001d36;
     --md-sys-color-outline: #79747e;
@@ -115,6 +117,8 @@
     --md-sys-color-on-error-container: #f9dedc;
     --md-sys-color-tertiary: #efb8c8;
     --md-sys-color-on-tertiary: #492532;
+    --md-sys-color-tertiary-container: #633b48;
+    --md-sys-color-on-tertiary-container: #ffd8e4;
     --md-sys-color-primary-container: #004a77;
     --md-sys-color-on-primary-container: #d0e4ff;
     --md-sys-color-outline: #938f99;
--- a/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-mode-commands.ts
+++ b/packages/dashboard/src/pages/cluster-commands/clusters/device-energy-management-mode-commands.ts
@@ -1,0 +1,162 @@
+/**
+ * @license
+ * Copyright 2025-2026 Open Home Foundation
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import "@material/web/button/outlined-button";
+import { css, html, nothing, type CSSResultGroup } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import { handleAsync } from "../../../util/async-handler.js";
+import {
+    DEVICE_ENERGY_MANAGEMENT_MODE_CLUSTER_ID,
+    decodeChangeToModeResult,
+    deviceEnergyManagementModeInfo,
+    type ChangeToModeResult,
+} from "../../../util/device-energy-management-mode.js";
+import { errorText } from "../../../util/error-text.js";
+import { BaseClusterCommands } from "../base-cluster-commands.js";
+import { registerClusterCommands } from "../registry.js";
+
+const CLUSTER_ID = DEVICE_ENERGY_MANAGEMENT_MODE_CLUSTER_ID;
+
+@customElement("device-energy-management-mode-cluster-commands")
+class DeviceEnergyManagementModeClusterCommands extends BaseClusterCommands {
+    @state() private _selectedMode: number | null = null;
+    @state() private _busy = false;
+    @state() private _result?: { mode: number; result: ChangeToModeResult };
+    @state() private _error?: string;
+
+    private async _changeToMode() {
+        const node = this.node;
+        const endpoint = this.endpoint;
+        const mode = this._selectedMode;
+        if (mode === null) return;
+        this._busy = true;
+        this._error = undefined;
+        this._result = undefined;
+        try {
+            const response = await this.client.deviceCommand(node.node_id, endpoint, CLUSTER_ID, "ChangeToMode", {
+                newMode: mode,
+            });
+            if (!this.isSameContext(node, endpoint)) return;
+            this._result = { mode, result: decodeChangeToModeResult(response) };
+        } catch (err) {
+            if (this.isSameContext(node, endpoint)) this._error = errorText(err);
+        } finally {
+            if (this.isSameContext(node, endpoint)) this._busy = false;
+        }
+    }
+
+    override render() {
+        if (!this.node || this.cluster !== CLUSTER_ID) return nothing;
+        const info = deviceEnergyManagementModeInfo(this.node.attributes, this.endpoint);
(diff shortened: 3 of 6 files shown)


Apollon77:
Reviewed and pushed a fix commit to this branch (rebased onto current `main`, CHANGELOG conflict resolved). All six Copilot comments are correct; two of them are blockers and they cluster into two design fixes rather than six patches.

**1. The response was decoded from the wrong wire format.** I traced this end to end rather than taking it on faith: `WebSocketControllerHandler.ts:1262` → `#convertCommandDataToWebSocket` → `convertMatterToWebSocketNameBased`, whose struct branch emits `matterNameToWireField(...)` — so a `ChangeToModeResponse` arrives as a bare, name-keyed `{status, statusText}`, with no outer envelope and no numeric keys. (Attributes go the other way, through `convertMatterToWebSocketTagBased`, which is why `tagField` is right for `SupportedModes` and wrong here. `door-lock.ts:decodeUserResponse` is the in-repo precedent: `pickNumber(obj, "userName")` etc.)

So `field(response, 0)` was always `undefined` and the `?? 0` fallback reported every device rejection as `Success` — the panel would render a green `ChangeToMode(2) → Success (0)` while `CurrentMode` never moved. That is the panel's entire stated reason for bypassing `sendCommand()`, dead on every non-success path. It decodes `status`/`statusText` by name now, and a response carrying no status throws rather than defaulting to Success.

The four fixtures were tag-keyed, so they certified the defect — including one literally named *"treats a missing status as Success"*. They now use the response shape, plus a case pinning the tag-keyed form as unreadable.

**2. Panel state was not keyed to the node.** `_selectedMode`, `_result`, `_error` and `_busy` all survived a navigation, and `_busy` reset only under `isSameContext`, which `cluster-commands/AGENTS.md` allows only for the EP0-only ICD panel with its `updated()` force-reset — neither applies to an application cluster on an arbitrary endpoint. Three reachable failures: an invoke in flight during a navigation left the button permanently disabled; a mode selected on node A could be sent to node B (whose `SupportedModes` may not even contain it) while A's stale result box was still displayed under B; and an A→B→A return adopted the pre-navigation outcome as if fresh. Fixed with a `willUpdate` context reset plus a generation token, the way `door-lock-commands.ts` does it.

Also in the commit:

- **Clicking the button without touching the dropdown sent nothing.** The select rendered `this._selectedMode ?? info.currentMode ?? supportedModes[0]?.mode`, but the handler read only `_selectedMode`, which stays `null` until a `change` event — and re-picking the already-shown option does not fire one. Silent return, no command, no feedback. The handler now receives the resolved value.
- **`--md-sys-color-tertiary-container` / `-on-tertiary-container` are defined nowhere in `public/index.html`**, for either theme, so both declarations were invalid and the success box rendered as plain text. I added the two tokens to both theme blocks rather than swapping the panel to another pair — `access-control-commands.ts:343,376` uses the same broken pair, so this repairs that too.
- `role="status"` / `role="alert"` on the outcome and error boxes.
- The mode picker uses `md-outlined-select` / `md-select-option` like the other panels instead of a bare `<select>`.

**One spec correction you should know about:** the PR body says the panel "shows `StartUpMode`/`OnMode` when the device exposes them", but `device-energy-management-mode.element.js` overrides both with `conformance: "X"` — disallowed on this cluster, unlike Mode Base where they are `O`/`DEPONOFF`. Those rows could never appear on a conformant device, so I removed them and the two attribute reads. Worth fixing the description too.

Apollon77:
@lboue the description needs one correction, and it is a substantive one rather than a wording nit.

**`StartUpMode` and `OnMode` are disallowed on this cluster.** The description says the panel *"Shows the current mode (resolved to its label), and `StartUpMode`/`OnMode` when the device exposes them"*, and cites `device-energy-management-mode.element.ts` as the source. That file actually overrides both:

```js
Attribute({ name: "StartUpMode", id: 2, conformance: "X" }),
Attribute({ name: "OnMode", id: 3, conformance: "X" }),
```

`conformance: "X"` means disallowed — unlike Mode Base, where they are `O` and `DEPONOFF` respectively (the cluster also pins the `DEPONOFF` feature to `X`). So a conformant DeviceEnergyManagementMode server never exposes either, and those two rows could never render. I removed them and the two attribute reads in the commit I pushed.

Suggested replacement for that bullet: *"Shows the current mode resolved to its label. `StartUpMode`/`OnMode` are `conformance: X` on this cluster — unlike the Mode Base parent — so they are deliberately not read."*

Two more worth reflecting in the description, since they change what the panel does:

- The bullet about sending `ChangeToMode` via `client.deviceCommand()` *"so the `ChangeToModeResponse`'s `Status`/`StatusText` is decoded and shown"* was true in intent but not in effect — the response was decoded by TLV tag while command responses arrive name-keyed, so every rejection displayed as `Success`. It works now, but if you keep that sentence it should not read as if it always did.
- The last bullet says the panel *"isn't registered with `renderWhenOffline`"* because it issues an active command. That is still accurate and still the right call.

Apollon77:
Both new Copilot comments are correct.

**MfgCode dropped from the tag label — fixed and pushed.** `MfgCode` is the tag's namespace, so the same `Value` means different things under different vendors; resolving a manufacturer tag against the standard table rendered vendor tag `0x0000` as "Auto". It now reads `Mfg 0x1234 tag 0x0000`, reusing the `formatHex` helper and the same shape `describeSemanticTag()` already uses for exactly this problem in `semantic-tags.ts`. Test covers a manufacturer tag whose value collides with a standard one.

**StartUpMode/OnMode in the description — code already done, description still needs @lboue.** I removed both rows and their attribute reads in an earlier commit, for the reason Copilot gives: the cluster overrides both with `conformance: "X"`. So the code no longer advertises them; it is only the PR body that still does. My earlier comment above has suggested replacement wording.

445/445 green.

lboue:
Thank you for the corrections you’ve made

Apollon77:
This round's finding is the PR description again, not the code — and Copilot's conclusion matches mine, so there is nothing to change in the implementation.

To restate it once so it is unambiguous: `StartUpMode` (id 2) and `OnMode` (id 3) are overridden with `conformance: "X"` on this cluster in `device-energy-management-mode.element.js`, i.e. **disallowed**, unlike the Mode Base parent where they are `O` / `DEPONOFF`. A conformant DeviceEnergyManagementMode server never exposes them. I removed both the decoding and the rows in an earlier commit for that reason; the review is flagging that the PR body still advertises them.

@lboue this needs your edit — replace the `StartUpMode`/`OnMode` bullet in the description with something like:

> Shows the current mode resolved to its label. `StartUpMode`/`OnMode` are `conformance: X` on this cluster — unlike the Mode Base parent — so they are deliberately not read.

I have left the thread open rather than resolving it, since the change it asks for is in the description and I have not made it.

Nothing else this round; the manufacturer-tag namespace fix from the previous round is in. 445/445 green.

Apollon77:

**Correction — `StartUpMode` and `OnMode` are not supported on this cluster.** The description says the panel *"Shows the current mode (resolved to its label), and `StartUpMode`/`OnMode` when the device exposes them"*, citing `device-energy-management-mode.element.ts`. That file overrides both:

```js
Attribute({ name: "StartUpMode", id: 2, conformance: "X" }),
Attribute({ name: "OnMode",      id: 3, conformance: "X" }),
```

`conformance: "X"` is *disallowed* — unlike the Mode Base parent, where they are `O` and `DEPONOFF` (this cluster pins the `DEPONOFF` feature to `X` as well). A conformant server never exposes them, so those rows could never render. Both the rows and their attribute reads were removed during review. Read that bullet as: *shows the current mode resolved to its label; `StartUpMode`/`OnMode` are `conformance: X` here and are deliberately not read.*

**Clarification — the response decoding.** The bullet about sending `ChangeToMode` via `client.deviceCommand()` *"so the `ChangeToModeResponse`'s `Status`/`StatusText` is decoded and shown"* describes the intent, but as originally written the response was read by TLV tag while command responses arrive name-keyed, so every device rejection displayed as `Success`. That is fixed; the sentence is accurate now, just not retroactively.

**Two behaviours added during review:** a manufacturer mode tag keeps its vendor namespace (`Mfg 0x1234 tag 0x0000` rather than resolving to the standard table's "Auto"), and per-node state is reset on navigation with a generation token, so a mode selected for one device cannot be sent to the next and an in-flight invoke cannot wedge the button.

The last bullet — that the panel is deliberately **not** registered with `renderWhenOffline` because it issues an active command — is still accurate.

Apollon77 (approved):

Issue #1059, opened 2026-09-13
Nested Aggregator endpoint's children are not recognized as bridged devices


lboue:
### Matter Server version

matter-python-client 1.4.1-alpha.2-20260830-4629759 (python_client), matter-server package from the same monorepo checkout (packages/matter-server, `@matter/main` 0.17.9)

### How do you run the Matter Server?

Standalone npm / node

### Node.js version

v24.18.0

### Operating system

Linux (Ubuntu-based), x64

### Description

**What I did**: Commissioned `matterbridge --demo-devices --profile Demo` (Matterbridge 3.10.10) to a local matter-server instance, then inspected the resulting `MatterNode` structure via the Python client (`python_client/matter_server/client/models/node.py`).

**What I expected**: The two nested `Aggregator` bridged endpoints in the demo set (endpoint 606 "Generic Switch" group and endpoint 1102 "Aggregator" group) should have their own children (6061-6064, 11021-11024) recognized as independently bridged devices, resolvable to that Aggregator as their parent.

**What happened**: `MatterNode.update()` (node.py, around the loop building `_composed_endpoints`) explicitly skips any endpoint whose device type list contains `Aggregator`:

```python
if Aggregator in endpoint.device_types:
    # ignore Bridge endpoint
    # (as that will also use partsList to indicate its child's)
    continue
```

This is correct for the top-level bridge aggregator (endpoint 1), but it also fires for a *nested* Aggregator endpoint (one that is itself a bridged device — carries `BridgedNode` alongside `Aggregator` — with its own children). Endpoint 1102 (`types=['BridgedNode', 'PowerSource', 'Aggregator']`) has PartsList `[11021, 11022, 11023, 11024]`, but because it carries `Aggregator`, that PartsList is discarded entirely instead of being recorded. Its four children:

- are not registered as `_composed_endpoints` (the `continue` skips it)
- do not carry their own `BridgedNode` device type (only endpoint 1102 does)
- so `get_compose_parent()` returns `None` and `is_bridged_device` is `False` for all four

They end up indistinguishable from a plain, non-bridged, top-level node endpoint — no relationship to endpoint 1102 is recoverable anywhere in the client model.

[...]

PR #1060, opened 2026-09-13
Recognize bridged devices nested behind another Aggregator endpoint

lboue:
## Type of change

## Description

`MatterNode.update()` skips any endpoint carrying the `Aggregator` device type when building the composed-device parent map, discarding its `partsList` entirely. That's correct for the top-level bridge aggregator, but it also silently drops the relationship for a *nested* Aggregator endpoint — one that is itself a bridged device (`Aggregator` + `BridgedNode` on the same endpoint) with its own further bridged children. Those children end up with no compose parent and no `BridgedNode` of their own, so `is_bridged_device` is `False` for them and the relationship to their aggregator is unrecoverable anywhere in the client model.

Fix for https://github.com/matter-js/matterjs-server/issues/1059

Changes:

- Track `Aggregator` `partsList` members in a new, separate `_bridge_parents` map instead of discarding them
- Kept separate from `_composed_endpoints` so they don't get merged into a single device the way composed-device parts do (they're independent bridged devices, not sub-parts of the aggregator).
- Add `MatterNode.get_bridge_parent()`.
- Extend `MatterEndpoint.is_bridged_device` to also return `True` when the endpoint has a bridge parent.
- Extend `MatterEndpoint.device_info` to fall back to the bridge parent's info when the endpoint has neither its own `BridgedDeviceBasicInformation` nor a compose parent.

Backward compatible for the common case: a top-level bridged device's own parent (the node's root Aggregator) now also gets tracked via `_bridge_parents` instead of being silently discarded, but since that aggregator endpoint typically isn't itself bridged, it still resolves to the same node-level identity as before.

## Backing evidence

- Related issue: homeassistant/core#182091 — contains the complete `--log-level debug` server log, the matterbridge log, and the full untrimmed raw node dump (147 endpoints) this analysis is based on, captured from a fresh `matterbridge --demo-devices --profile Demo` commission (endpoints 606 and 1102, each an Aggregator + BridgedNode with 4 bridged children of their own).
- [x] This fix/change is backed by a **complete log file** — attached here or in the linked issue (not just a few quoted lines).

## Checklist

- [x] Tests added or updated to cover the change — none added in this repo (no existing unit tests cover `MatterNode`/`MatterEndpoint` parsing here; `python_client/tests/` is integration/mock-server style). Coverage currently lives on the Home Assistant side (`test_bridged_device_behind_nested_aggregator`, homeassistant/core PR TBD). Open to adding one here in whatever shape you'd prefer.
- [x] `npm test` passes — not run as such (that's the JS/`packages/*` suite, untouched by this change). The Python client suite (`npm run python:test`) passes: 76/76.
- [x] `npm run format-verify` and `npm run lint` pass — not run as such (JS-side, no files touched there). The Python-side equivalents pass: `npm run python:lint` (ruff) all checks passed, `npm run python:typecheck` (mypy) 0 issues across 188 files.


python_client/tests/test_bridge_topology.py
```diff
@@ -0,0 +1,110 @@
...
+            0: {
+                _DEVICE_TYPE_LIST: [{"0": _ROOT_NODE, "1": 1}],
+            },
+            1: {
+                _DEVICE_TYPE_LIST: [{"0": _AGGREGATOR, "1": 1}],
+                _PARTS_LIST: [100],
+            },
+            100: {
+                _DEVICE_TYPE_LIST: [
+                    {"0": _AGGREGATOR, "1": 1},
```
Apollon77, line 63: nooo ... this must not have an aggregator! only bridged node is allowed here

Apollon77: And yes it is allowed to have multiple endoints under one bridged node and such (composed devices) ....

Apollon77 (changes requested):

Luligu:
Hi,
image

is a normal and spec compliant use.

It allows to give a name to children.

And pass of course chip test device composition.

python_client/tests/test_bridge_topology.py
Apollon77: ok it is fine, i was wrong, verified

Apollon77:
I will prepare a replacement PR for this because found more things for this case. So this gets closed likely as synced with @lboue. but stays open until replacement is there

Apollon77:
Closing in favour of #1077, which carries the fix for #1059.

Your report and the node dump are what made the case reproducible, and the nested-aggregator shape you identified is now part of the test bridge device both integration suites commission, so it stays covered.

#1077 resolves the parent differently: instead of writing every Aggregator `partsList` entry into the map, it picks the closest parent — of the endpoints listing a given endpoint, the one that lists none of the others. Your version depended on endpoints being visited in ascending order for the nested aggregator to win, which held for the Matterbridge device but is not guaranteed.

It also does not take the Aggregator's device info for a bridged device that reports none of its own: with two such children under one aggregator they would share `nodeLabel`, `serialNumber` and `uniqueID`, which merges them into one device in a consumer's registry. Those endpoints now get `None` and a log line, and `get_bridge_parent()` carries the relation instead.

Issue #1061, opened 2026-09-13
Thread graph hides a reachable native Thread router observed by one Matter device


gbertsch:
### Matter Server version

matter-server 1.4.0 / matter.js 0.17.9; Home Assistant Matter Server app 9.2.0 with beta enabled.

### How do you run the Matter Server?

Home Assistant Add-on

### Operating system

Home Assistant OS 18.3.rc1, x86_64; Home Assistant Core 2026.9.2.

### Description

I have two Shelly 1PM Gen4 devices running native ESPHome over Thread (not Matter). One was missing from the Matter dashboard's Thread graph, although searching its exact extended address opened its External Device details and showed one commissioned IKEA light as its observer. The other Shelly appeared as an external router.

An independently refreshed OTBR device collection reported both Shellys as routers. I expected the graph to represent the missing router, or indicate that its evidence was considered uncertain, instead of silently excluding its icon.

The observed conditions were:
1. A native non-Matter Thread router appears in one online commissioned Matter device's neighbour table.
2. That observer has more than one neighbour.
3. Exact-address search opens the external router's details, but its node is absent from the graph.

This describes my observed setup, not a claim that every network with these conditions reproduces it.

### Investigation
AI assisted the investigation and drafting. The relevant source is:
https://github.com/matter-js/matterjs-server/blob/v1.4.0/packages/dashboard/src/pages/network/thread-graph.ts#L364-L382

> The frontend's single-observer branch hides an external device when it has exactly one observer and that observer has more than one neighbour. A local, display-only experiment removed that branch, after which both Shelly router icons appeared. The remaining offline-observer and border-router filtering was left unchanged.

This suggests a false-negative in the display heuristic. It does not establish that every single-observer entry is current. Showing an uncertainty indicator may be safer than removing stale-data protections altogether. No commissioning or routing failure was established, and the experiment did not change network credentials, pairing, firmware, or backend behaviour.

### Supporting log evidence
The accompanying `matter-thread-map-2026-09-13.redacted.log` is a continuous excerpt of the existing debug log, September 13, 2026, 08:59:00.017–09:32:59.999 (local time, America/Toronto). It retains all 4,652 lines in that window, including unrelated messages. No fresh reproduction was performed.

[...]

PR #1073, opened 2026-09-17
fix(dashboard): keep an external Thread device visible when Thread diagnostics confirm it

Apollon77:
Addresses #1061 (the reported symptom; see "Not covered" below for what is left).

## Problem

The Thread graph hides an external (non-commissioned) Thread device when exactly one Matter node reports it in its neighbor table and that node has other neighbors. The filter is there to drop single-source ghosts left in a stale neighbor cache. It only ever looked at neighbor tables, never at the Thread diagnostics the graph already holds, so a device confirmed by a Border Router was still dropped.

Effect: a native (non-Matter) Thread router seen by exactly one commissioned device is missing from the graph, while the dashboard search still finds it and opens its details panel — the two views disagree about whether the device exists.

The reporter's log shows the full chain: ext address `b00000000000000b` appears in one neighbor table (node `@1:62`, three neighbors), and the same address appears in the MeshCoP diagnostics batch as `rloc16 0xcc00`.

## Change

A diagnostics record for the same extended address is evidence from a source other than the observing node, so it lifts both stale-cache filters; the device then follows the "hide offline nodes" toggle like any other node.

The record must be trustworthy to count:

- **Complete batch only.** A batch with `partialReason` set is an aborted query, so an entry in it can predate the failure. This matches `_brBatch`, which already rejects partial batches as evidence for Border Routers.
- **Same Thread network only.** Matching on the extended address alone would let a batch for network A vouch for a device on network B. Every other diagnostic lookup in this file is keyed per extended PAN ID.
- **Age is not re-judged here.** The server already bounds it with the diagnostics cache TTL; a second freshness rule in the dashboard would drift from it.

The decision moved out of the Lit component into `shouldHideExternalDevice` in `ws-client`, next to the rest of the topology derivation, so it is testable without a DOM. `network-details.ts` had a byte-identical copy of the diagnostics lookup; both call sites now share `findDiagnosticRecordByExtAddress`.

`TopologySourceNode` gains an optional `available`, which the rule reads to decide whether an observer can still confirm what it reported. Existing callers are unaffected — the field is optional and an absent value means online.

## Tests

13 tests in `TopologyUtilsTest.ts`, each added branch mutation-checked: dropping the `kind === "br"` clause, dropping the diagnostics clause, deleting the all-observers-offline return, accepting a partial batch, accepting a batch from another network, `&&` → `||`, `> 1` → `> 0`, `!== 1` → `!== 2`. Every one of those fails at least one test.

One guard is deliberately left uncovered: `observer !== undefined` on the single-observer branch cannot be reached, because an online observer already implies the node is in the map.

`npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` all pass.

## Also fixed here

Found while working on the above, and fixed in this PR:

- **`selectBySearch` offered devices the graph hides.** It claimed only visible nodes are selectable but skipped the check for external devices, which is how the reporter reached a details panel for a device with no icon. It now calls `shouldHideExternalDevice` with the same inputs the graph uses. No test: the method needs a DOM and the dashboard suite is util-only.
- **`network-details.ts` counted an observer as online only on `available === true`**, where the rest of the code treats a missing flag as online. Both now call a shared `isObserverOnline`, which also rejects an id with no node behind it.
- **The server served diagnostics batches it calls stale.** `get_thread_diagnostics` without an `ext_pan_id` answered from `listCached()`, which returned the whole cache; the TTL applied only to the single-network path and nothing evicts an entry. It now withholds anything past the TTL, including the snapshots of a query still in flight, which carry nodes. A terminal partial still survives at any age: it carries the reason a query failed, and the panel that renders that reason shows nothing without the batch.
- **The dashboard reinstated batches the server had dropped.** The network view replayed every entry of the accumulating `MatterClient.threadDiagnostics` map on each update, so an update for one network brought back an expired batch for another. The client now offers a typed listener (mirroring the existing WebRTC one) and the view applies only the batch that arrived.

## Not covered

- `NetworkTopologyService` still publishes every external device with a live edge, deliberately. It has no MeshCoP diagnostics — the evidence a ghost is checked against — so filtering there would drop exactly the devices this PR restores. The wire carries the evidence; visibility stays a display decision, and the class note now says so.
- `matter-network-view.ts` uses the `void` operator on a fire-and-forget promise, and `BorderRouterStore.reset()` has no caller although `client.ts` clears the client-side map on `start_listening` for exactly the reason the store would need it. Both pre-existing and untouched here.


Apollon77:
Addressed the review in 825463de, plus the three items the PR body listed as not covered.

**Review findings.** Both comments are one root cause: the code split "find a record" from "judge the record", so the finder could return a record that does not qualify and shadow one that does, and the judge had to guess when the device's network was unknown. Rather than guard each symptom, `corroboratesExternalDevice` is gone and `findCorroboratingDiagnostics` searches for a qualifying record:

- complete batches only (`partialReason` unset);
- the device's own Thread network, when the observing node reports an extended PAN ID;
- otherwise only while a single network reports that address — with candidates in two networks there is nothing to decide between them, so the device is not corroborated.

This removes the first-match problem (all batches are scanned) and the accept-unknown-network problem (a missing PAN ID no longer means "any network will do"). Three added tests, each mutation-checked: dropping the same-network filter, accepting an ambiguous match, and accepting a partial batch each fail a test.

**Items from the PR body.**

1. `NetworkTopologyService` — left publishing every external with a live edge, deliberately. It has no MeshCoP diagnostics, which is the evidence a ghost is checked against, so filtering there would drop exactly the devices this PR restores. The wire carries the evidence and visibility stays a display decision; the class note now records that.
2. `selectBySearch` — fixed. It offered external devices the graph hides, which is how the reporter reached a details panel for a device with no icon. It now calls `shouldHideExternalDevice` with the same inputs the graph uses, so the two cannot disagree. No test: the method needs a DOM and the dashboard suite is util-only.
3. `network-details.ts` — fixed. It counted an observer as online only on `available === true`, where the rest of the code treats a missing flag as online. Both now call the shared `isObserverOnline`, which also rejects an id with no node behind it. Three tests, mutation-checked against `node?.available !== false`.

`npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` pass on the merged tip.

Apollon77:
Both comments were right; addressed in ac5bd549.

**Stale batches reaching a client that already holds them.** Correct, and it named the gap the previous commit left open. `MatterClient.threadDiagnostics` accumulates for the life of the connection, and the network view replayed every entry of that map on each update, so an update for network B reinstated an expired network A in the store. Taking the second of the two options offered: the client now exposes a typed `thread_diagnostics_updated` listener — same shape as the existing WebRTC one, no public signature change — and the view applies only the batch that arrived. A refresh rebuilds the store from the server snapshot, which is now TTL-filtered, so a dropped entry stays dropped. Mutation-checked: replaying the whole map to the listener fails the new `WsClientTest` case.

**In-progress snapshots.** Also correct, and worth stating plainly: my first test for this was vacuous. After a successful fetch the cache holds the final complete batch, so there was no streaming partial left to filter and emptying the reason set did not fail anything. Rewritten to hold a stream open and assert on the `in_progress` snapshot; the mutation now fails. `listCached` expires `in_progress` and `meshcop_no_responses_yet` with complete batches — they carry nodes, and an old one means a stream that never finished. Only terminal partials survive their TTL, since they carry the reason a query failed and the panel that renders it shows nothing without the batch.

**Scope in the PR description.** Fixed. The search-visibility change and the `available === true` change moved out of "Not covered" into a new "Also fixed here" section, together with the two server-side fixes made since. What remains under "Not covered" is `NetworkTopologyService`, which keeps publishing every external with a live edge on purpose — it has no MeshCoP diagnostics, so filtering there would drop exactly the devices this PR restores.

Gates on the merged tip: `npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` pass. One note for the record: two intermediate runs failed `Integration Test > Device Discovery`, which is the known mDNS flake; a clean rebuild plus three consecutive full runs are green.

Apollon77:
All three addressed in 162acca4.

**`rest_no_responses_yet` was missing.** A real bug, not a wording issue: that reason exists (`model.ts:190`) and is produced (`ThreadDiagnosticsService.ts:749`), so a REST in-flight snapshot survived the TTL — the case the filter was added to stop. The cause was writing a subset of a union out by hand, so rather than adding the missing string the set is now a `Record<ThreadDiagnosticsPartialReason, "streaming" | "terminal">` next to the union it classifies. A reason added without a kind is a compile error; verified by deleting a key and getting `TS2741`.

**The declaration split the class documentation from the class.** Correct, and fixed by the same move — the record now sits with the type, and the `ThreadDiagnosticsService` JSDoc is attached to the class again.

**Both doc texts called every partial a failed query.** Fixed in `CHANGELOG.md` and in `docs/websocket-api-schema-changelog.md`, each now naming the terminal reasons that survive the TTL and the three in-flight ones that expire with complete batches.

Behaviour is mutation-checked: flipping `in_progress` to `terminal` fails `listCached withholds a streaming snapshot past the cache TTL`.

Gates: `npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` pass.

Apollon77:
Round 5 addressed in 3f7beeee, and I ran a full adversarial review over the whole branch rather than only the latest change. It found a regression this branch had introduced, so this comment is mostly about that.

**The suppressed finding was right, and fixing it naively made things worse.** A client that already held a batch kept it forever — the store has no expiry and the Thread panel has no periodic refresh — so a graph open past the cache TTL went on drawing devices the server would no longer vouch for. That window is ours: before this branch such a device was not drawn at all.

A batch now carries `expiresInMs`, the lifetime remaining when the server sent it, on all three serialization paths. It is a duration rather than a deadline so each side measures on its own clock; optional and additive, so no schema bump.

My first implementation pruned the batch when that lifetime ran out. Our own review caught what that does: after one TTL the device disappears again and **nothing ever refetches** — `_refreshBorderRouters` runs only on a network-type change or `connections-updated`, and the server only refreshes when a client asks. A Thread view left open for an hour silently re-created the bug this PR exists to fix, and took the diagnostic mesh nodes, diagnostic edges and BR diagnostics panels with it.

So expiry now triggers a **refetch**, not a delete. The refreshed snapshot is authoritative both ways: a network that still reports comes back current, one that does not is omitted. Pruning is the fallback when the refetch fails, so losing the server degrades to distrusting old evidence rather than keeping it.

Also fixed from the same review, each with the failure it prevents:

- Both store maps now swap together. A throw mid-refresh previously left batches behind with their deadlines cleared — permanently un-expirable, reached through the error path of the code meant to prevent staleness.
- Deadlines moved to the monotonic clock. On the wall clock an NTP step or manual clock change moved every deadline at once, defeating the property the relative encoding was chosen for.
- A lifetime that is not a positive finite number is treated as no lifetime, and the timer delay is clamped. `NaN` off the wire, or a lifetime beyond the 32-bit `setTimeout` range, each produced an unbounded reschedule loop in the browser tab.
- The timer is re-armed on reconnect and never armed after teardown.
- An unknown device takes its graph label from the record that vouches for it, instead of from any batch naming its address. Label and visibility were reading contradictory evidence: a device could be hidden as uncorroborated while showing vendor text drawn from the very batch the visibility rule rejected.
- `listCached`'s JSDoc had been orphaned onto `remainingTtlMs` (the same slip as 162acca4), and its rationale was inverted: it credited `getOrFetch` with refusing stale batches, but `getOrFetch` refuses a streaming partial at any age.

New tests: 10 for store expiry, 3 for `serializeBatch`, 1 for `remainingTtlMs`. Mutation-checked — dropping the finite/positive guard, clearing deadlines before the swap, always emitting `expiresInMs`, inverting the prune comparison and stretching the lifetime each fail a test.

One documented claim I checked and rejected: that `expiresInMs` should be version-gated because schema 13 is unreleased. `docs/websocket-api-schema-changelog.md:22` records that 1.4.0 shipped schema 13, so a field added now genuinely has no version signal and "presence-detectable only" is correct.

Two things deliberately not done: three partial service test stubs still lack `remainingTtlMs` — they omit `listCached`/`getOrFetch` too and never emit, so adding one method would be inconsistent dead weight; and `network-details.ts:1292` still uses `available === true`, which I left rather than change behaviour in code this PR does not otherwise touch.

Gates: `npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` pass.

Apollon77:
Round 6 addressed in 6dd269f1, plus a type cleanup in 9a14ae36.

**Zero remaining lifetime.** Correct, and the direction matters: my guard grouped `0` with malformed input and cleared the deadline, so the batch with no lifetime left was the one the store kept forever. It is reachable exactly as described — `remainingTtl` clamps at zero and the event path serializes lazily at flush, so a send delayed past the TTL carries zero. Now only an absent lifetime means "never expires"; zero or negative yields a deadline already past, and the batch goes at the next check. Mutation-checked: restoring the `lifetime <= 0` grouping fails a test.

**The test that codified it** is split into three: zero drops, negative drops, non-finite is ignored. The non-finite case stays because JSON carries neither `NaN` nor `Infinity` — it guards against a malformed client, and arithmetic on such a value would spin the reschedule loop instead of expiring anything.

**`partialReason` wording.** Right, `in_progress`, `meshcop_no_responses_yet` and `rest_no_responses_yet` describe a query still running, not an aborted one. The doc now states what both cases share: neither is a complete view of the network.

**Separately**, a reviewer here asked why the new method used a plain millisecond number when this package types its timings with matter.js `Duration` — poll intervals, watchdogs, interaction timeouts all do. It was following the timing options of the file it sits in rather than the convention around it. `remainingTtl` now returns `Duration`, so the unit lives in the type rather than the name, and the conversion happens once at the boundary: `serializeBatch` unwraps with `Millis.of`, since the wire carries plain JSON numbers and `ws-client` depends on no matter.js package. The wire field is unchanged.

The service's own options (`cacheTtlMs`, `windowMs`, `firstBatchMs`, `debounceMs`, `restProbeTimeoutMs`) remain plain numbers. Converting them changes the service's public option shape and every caller that builds one, so that belongs in its own change rather than buried here.

Gates: `npm run format-verify`, `npm run lint`, `npm run build` and the full `npm test` pass.

Apollon77:
Round 7 addressed in e02373d4. Both findings were correct.

**Thread work ran from the Wi-Fi view.** The connection stays opted in to diagnostics events after the user switches views, so a late batch armed the timer and started Thread mesh collections for a graph that is not on screen — real radio traffic, not just a wasted tick. Scheduling and the timer callback now both require `networkType === "thread"`; the callback too, since the view can switch between arming and firing. `_handleConnectionsUpdated` directly below already applied that convention and I had missed it.

**Pruning after the await let spent evidence live for minutes.** The default WebSocket command timeout is five minutes (`client.ts:54`), so a stalled `get_thread_diagnostics` kept an expired batch corroborating external devices for that whole window, which is what `expiresInMs` exists to prevent. The callback now drops first and asks second: the deadline has passed, so the evidence is spent whatever the refresh returns, and the refresh that follows repopulates what is still current.

That reorder is a visible behaviour change worth naming: a device can now disappear briefly between the drop and the refresh response, where before it stayed on screen. Showing unsupported evidence seemed the worse of the two, and the refetch still prevents the "vanishes with nothing to bring it back" case from an earlier round.

Neither change has a test. `matter-network-view` needs a DOM and the dashboard suite is util-only, so these are verified by reading rather than by a test that fails without them — stating that rather than implying coverage.

Gates: `npm run format-verify`, `npm run lint`, `npm run build` and the dashboard suite (561/561) pass. One local full-suite run showed the known `matter-server` integration cascade (`Timeout waiting for error response to commission_with_code`, then the mDNS discovery tests); it reproduces on this machine without these changes, and the only file touched here is a browser component those tests never load. CI is the cross-check.

Apollon77:
Round 8 addressed in 2a3b06fc. All three findings are the same defect, so I fixed what produced it rather than the three sentences.

**The claim was wrong.** `meshcop_no_responses_yet` and `rest_no_responses_yet` are selected exactly when `acc.size === 0`, so two of the three streaming reasons carry no nodes at all. Saying they expire "because they carry nodes" was false for both.

The real reason holds for all three and does not depend on contents: such a batch reports the state of a query that was still running when it was taken, and that query finished or died long ago. The snapshot stops being a current account of the network whether or not any node had answered.

**Why it was wrong in three places at once:** the reasoning was written out three times — in the `listCached` JSDoc, in the schema changelog and in `CHANGELOG.md` — so one imprecision triplicated. This is the third round where the wording around `partialReason` needed correcting (rounds 4, 6, 8), which is a pattern in the prose rather than three unrelated slips.

So the rationale now lives once, next to `PARTIAL_REASON_KIND`, which is the thing that actually decides the behaviour. `listCached` points at it instead of restating it, and both changelog entries describe what a client receives rather than re-deriving the internals. The next correction to this rule has one place to land.

No behaviour change: documentation and comments only. Gates: `npm run format-verify`, `npm run lint`, `npm run build` and `npm test -w @matter-server/ws-controller` (407/407) pass.

Apollon77:
Both fixed in efa8fac4, and both were right for the reason that matters: the tests were vacuous.

`XP_HEX` and the `serializeBatch` fixture were digit-only hex, where `toUpperCase` and `toLowerCase` are identity. So "uppercases the extPanId for the wire" and "matches the network of a corroborating batch case-insensitively" each passed with the normalization they name deleted — they asserted nothing about casing at all.

The fixtures now contain hex letters (`1122334455667AAB`, `aabbccddeeff0011`), and I checked the fix rather than assuming it: deleting `batch.extPanIdHex.toUpperCase()` in `serializeBatch` now fails its test, and deleting the `toUpperCase()` on the network comparison in `findCorroboratingDiagnostics` now fails its test. Neither did before this change.

Test-only. Gates: `npm run format-verify`, `npm run lint`, `npm run build`, ws-client 120/120, ws-controller 407/407, dashboard 561/561.

Issue #1075, opened 2026-09-18
Feature request: Dashboard panel for the ServiceArea cluster (0x0150)


lboue:
### Description

The Dashboard has no dedicated panel for the **ServiceArea** cluster (0x0150). Nodes exposing this cluster (RVC / robotic vacuum-mop devices, and other "Room" service device types) currently fall back to the generic cluster view, which just lists raw attribute values with no visualization of maps, areas, or operating status — and no way to trigger area-related commands from the UI.

**Problem it solves:** anyone testing or operating a ServiceArea-capable device (e.g. a robot vacuum) through matterjs-server has to read/decode SupportedAreas, SupportedMaps, SelectedAreas, CurrentArea and Progress by hand instead of seeing them as an actual room/zone picker with live status.

**Use case:** I'm working with Matter RVC-type devices and want to select which rooms to clean and see per-area progress directly from the Dashboard, without a companion app. This would also let me test ServiceArea features that Home Assistant's Matter integration doesn't support yet, independently of HA — validating device behavior (area selection, skip, progress reporting) directly against the server before/without waiting on HA-side support.

**Proposed approach** — a `ServiceArea` panel, similar in spirit to the `DeviceEnergyManagementMode` panel added in #1050:

- **SupportedAreas** (`AreaStruct[]`) rendered as a selectable list/grid (name, floor number, map ID, icon/color if AreaInfo is present)
- **SupportedMaps** (`MapStruct[]`) used to group/filter areas when a device has multiple maps
- **SelectedAreas** shown as toggles/checkboxes, calling the `SelectAreas` command on change
- **CurrentArea** highlighted in the list
- **EstimatedEndTime** shown as a countdown/timestamp
- **Progress** (`OperationalStatusStruct[]`) shown per area — Pending / Operating / Skipped / Completed
- `SkipArea` command exposed when the Skip feature bit is supported

Cluster ID `0x0150`, Matter 1.3+ (Device Library spec §1.11). I don't have a specific bug/log to attach — this is a net-new capability request, not a regression.

### Confirmation

- [ ] If this request is about a current limitation or wrong behavior, I attached a log file showing it.
- [x] I have reviewed and understand everything in this request and can explain it in my own words ([AI policy](https://github.com/matter-js/matterjs-server/blob/main/AI_POLICY.md)).
- [x] I searched existing issues and this is not a duplicate.

lboue:
# Proposed technical implementation plan

[...]

PR #1076, opened 2026-09-18
Adds dashboard command support for the Matter Service Area cluster

lboue:
**Type of change:** ✨ Feature

**Description:**

Adds dashboard command support for the Matter **Service Area** cluster (`0x0150`), for the feature request in #1075.

This introduces a `service-area-cluster-commands` custom element that:
- Renders supported areas grouped by map (when the `Maps` feature is supported), or as a flat list otherwise
- Seeds the area checkboxes from the device's `SelectedAreas` attribute, so submitting a selection does not cancel a job that is already running
- Lets the user select one or more areas and invoke `SelectAreas`
- Shows a per-area "Skip" action to invoke `SkipArea`, offered when `AcceptedCommandList` lists the command and the area is `Pending` or `Operating`
- Displays the estimated end time as an absolute instant plus a countdown that ticks once a second while the estimate is still in the future
- Shows progress badges per area (Pending / Operating / Skipped / Completed), the time each area has run and its estimated time, and the current-area indicator
- Names an area by its `LocationInfo`, its `LandmarkInfo` (with the relative position tag) or its area type, falling back to the area id
- Reports command results/errors inline, guarding against stale responses if the panel context (node/endpoint/cluster) changes mid-invoke (`_invokeGeneration` tracking)

The supporting decoder lives in `packages/dashboard/src/util/service-area.ts` and is unit tested in `packages/dashboard/test/ServiceAreaUtilTest.ts`. The new element is registered in `cluster-commands/index.ts` alongside the other cluster command panels.

**Review round 1 (commit `a44e228`, pushed by a maintainer):**

Command availability now comes from `AcceptedCommandList` (`0xfff9`) rather than from the feature bits. `SkipArea`'s conformance is `[CurrentArea | Progress]`, so the `PROG` bit neither implies the command nor is required by it: gating on `PROG` hid Skip on devices that accept it and offered it on devices that answer `UNSUPPORTED_COMMAND`. When a device reports no `AcceptedCommandList`, the panel falls back to the conformance rule. Skip also covers `Pending` areas, which are the queued ones a user most wants to drop.

`SelectedAreas` (`0x2`) is read. `SelectAreas` replaces the whole selection, so a panel whose checkboxes always started empty cancelled the areas a running job was working on as soon as the user pressed the button.

`ProgressStruct.Status` is carried as the numeric enum and the label is derived at render, so renaming a label cannot change which areas are skippable or reach a CSS class name. `LandmarkInfo` and `AreaType` are decoded, so an area a device identifies by landmark alone no longer renders as "Area 3". The countdown is driven by an interval cleared in `disconnectedCallback`, and `formatDuration`/`formatEpochTime` are reused instead of a local formatter.

`packages/dashboard/src/pages/cluster-commands/AGENTS.md` records why this panel's conditional `_busy` reset is safe on a cluster that is not endpoint-0-only.

**Related issue:** #1075

**Backing evidence:**

- [ ] This fix/change is backed by a complete log file — attached here or in the linked issue (not just a few quoted lines).

This is a net-new capability, not a regression, so there is no log to attach. It has not yet been exercised against a physical RVC device; the `AcceptedCommandList` gating and the landmark naming are the parts that most need that test.

**Verification:**

```
$ npm run format-verify
All matched files use the correct format.
Finished in 318ms on 287 files using 14 threads.

$ npm run lint
> oxlint --type-aware
(no findings)

$ npm run build
  ✓ Type check (.25s)
  ✓ Transpile (.079s)
created dist/web/js in 2.7s

$ PRIMARY_INTERFACE=en0 MATTER_MDNS_NETWORKINTERFACE=en0 npm test
  ✓ ESM 294/294 (1m 1s)
```

The dashboard's own suite covers the new decoder with 41 assertions. Five of them were mutation checked: removing the `AcceptedCommandList` gate, narrowing Skip back to `Operating` only, removing the conformance fallback, not decoding `SelectedAreas`, and pointing the landmark lookup at the wrong namespace each turn the suite red.

**Checklist:**

- [x] I understand the code I am submitting and can explain how it works (AI policy)
- [x] `npm test` passes
- [x] CHANGELOG updated


Apollon77 (approved):

Apollon77:
Thanks for this — the decoder split and the stale-response guard were both right, and the panel matched the `DeviceEnergyManagementMode` shape closely. I ran a full review pass over it and, rather than leaving you a list of separate findings to work through, pushed the changes to your branch as `a44e228`. Please read it before we merge: it is your PR, and under our AI policy you need to be able to explain what is in it. If you disagree with any part, say so and we change it back.

Two of the findings were behaviour bugs, and both came from the same root cause: the panel decided what it could do from decoded display state instead of from what the device declares.

**`SkipArea` was gated on the wrong thing.** The check was `features.progressReporting && progress?.status === "Operating"`. In the data model `SkipArea` has conformance `[CurrentArea | Progress]`, so the `PROG` feature bit neither implies the command nor is required by it. A device with `CurrentArea` and no `PROG` accepts `SkipArea` and got no button; a `PROG`-only device got a button whose invoke returns `UNSUPPORTED_COMMAND`. The panel now reads `AcceptedCommandList` (`0xfff9`), which is what four other util modules in this package already do (`media-playback.ts`, `camera.ts`, `door-lock.ts`, `energy-evse.ts`), and falls back to the conformance rule when a device reports no list. Skip is also offered for `Pending` areas now — those are the queued ones a user actually wants to drop, and the spec constrains `SkippedArea` to the selected set, not to the running area.

**`SelectedAreas` (`0x2`) was never read.** The checkboxes always started empty, but `SelectAreas` *replaces* the whole selection. Someone looking at a robot currently cleaning {Kitchen, Hall} would tick "Bedroom", press the button, and silently cancel Kitchen and Hall. The checkboxes now seed from `SelectedAreas` until the user edits them, and hand control back to the device after a successful invoke.

The rest is smaller:

- `ProgressStruct.Status` is kept as the numeric enum with the label derived at render. Before, the decoded label string was both the control predicate and the CSS class, so renaming a label would have silently removed the Skip button, and an out-of-range enum produced `class="progress-unknown(7)"`.
- `LandmarkInfo` (`AreaInfoStruct` field 1) is decoded. `LocationInfo` and `LandmarkInfo` are both nullable and the spec requires one of them, so a device naming its areas by landmark rendered a list of "Area 1", "Area 2", "Area 3".
- The countdown now ticks. It was computed from `Date.now()` at render time, and nothing re-renders on a clock — `tickContext` only fires on `nodes_changed` / `server_info_updated`. Since `EstimatedEndTime` is `X Q`, the device need not report changes at all, so the label could sit frozen for a whole job. There is now a 1 s interval that runs only while the estimate is in the future and is cleared in `disconnectedCallback`, and the absolute instant is shown next to it via the existing `formatEpochTime`.
- `formatDuration` from `util/duration.ts` replaces the local minute/second formatter, `floorNumber` / `totalOperationalTime` / `estimatedTime` are rendered instead of only decoded, and the `md-checkbox` event handler narrows with `instanceof` instead of casting to `HTMLInputElement` (which it is not — it is a `LitElement` that happens to expose `.checked`).
- Unit tests grew to 41 assertions. The FeatureMap test previously only used `0b111` and "absent", which cannot detect two bits being swapped; there is now one fixture per bit. I mutation checked five of the new guards — each turns the suite red when reverted.

Gates are green: `npm run format-verify`, `npm run lint`, `npm run build`, and `npm test` (294/294). I updated the PR description to match what now ships.

One thing I cannot check from here: none of this has run against a physical RVC. The `AcceptedCommandList` gating and the landmark naming are the parts most likely to behave differently against real hardware, so if you can point your device at it and report back, that is the last thing I want before merging.

lboue:
Thanks for the thorough review and for pushing `a44e228` directly, the explanations on `AcceptedCommandList` gating and on `SelectedAreas` seeding are very clear. I'll go through the commit and let you know if I disagree with anything.

I have real devce available (SwitchBot), so I'll test the panel against it and report back here. To cover more than one stack, I'll also run it against:

- the SDK `rvc-app` example, as a second reference implementation;
- my Matterbridge Service Area implementation, to build the cases hardware won't give me: landmark-only area names, with and without the `Maps` feature, `CurrentArea` without `ProgressReporting`, and edge values for `EstimatedEndTime` and progress status.

I'll post the device model and firmware, the `FeatureMap` and `AcceptedCommandList` of each target, and an expected-vs-observed summary, with the two points you flagged (`AcceptedCommandList` gating and landmark naming) checked first, on the physical device. I'll clearly separate what was seen on real hardware from what was seen on simulators.

lboue:
**Testing**

image

lboue:
**Testing**

image

image


image

lboue:
Hardware test report for the ServiceArea panel, on a real SwitchBot K11+ (native Matter 1.4.1,).

**Summary**

| Check | SwitchBot K11+ (native, real hardware) | Roborock Q10 S5+ via Matterbridge (my implementation, in development) |
|---|---|---|
| FeatureMap | `ProgressReporting` + `Maps` | `ProgressReporting` + `Maps` |
| AcceptedCommandList | `[0, 2]` | `[0, 2]` |
| Areas grouped by map | ✅ | ✅ |
| Checkboxes seeded from `SelectedAreas` | ✅ [confirm: areas were already selected on the device] | ✅ [same] |
| `SelectAreas` → `Success (0)` inline | [result] | ✅ |
| Skip on a selected area with no `Progress` entry (`a44e228`) | ❌ `InvalidInMode (2)`: "the skipped area does not match the current area" | not reported |
| Skip on a `Pending` area while no area is `Operating` (`a44e228`) | ❌ `InvalidInMode (2)` | not reported |
| Skip on the `Operating` / current area | [result] | not reported |
| Skip button after `0c06a8a` | [confirm: shown only when the invoke can succeed] | hidden while no area is `Operating` |
| Success banner colour | fixed in `0c06a8a` | fixed in `0c06a8a` |

**What I found with `a44e228`:** the Skip button was shown on every selected area, but the K11+ refuses it in the two cases above, both with `InvalidInMode`. Declaring `SkipArea` in AcceptedCommandList was therefore not enough on this device: the button was offered where the invoke could only fail (screenshot attached).

**What I changed in `0c06a8a`:**
- with no `Progress` entry, Skip is offered only for `CurrentArea` (previously any selected area);
- a `Pending` area counts as skippable only once some area is `Operating`;
- the success banner now uses the shared `--success-color` token, because the previous tertiary-container background (dark maroon in this theme) looked like the error state.

**Open question:** this is one device, and it partly narrows the `Pending` case you added in `a44e228`, which follows the spec's constraint on `SkippedArea`. I don't know whether other firmwares accept Skip on a `Pending` area earlier. If you'd rather not encode this device's behaviour, the alternative is to keep the button visible whenever the command is declared and rely on the inline error. Happy to go either way.

**Not covered:** [e.g. a second native device, `CurrentArea` without `ProgressReporting`]. The Matterbridge column is only there to exercise the rendering (map grouping, area names, `Pending` badges, checkbox seeding): my implementation is still in development and Skip does not work there yet, so I'm not reporting any Skip result from it.

PR #1079, opened 2026-09-19
Add Aqara FP400 custom clusters

RAR:
## Type of change

## Description

Adds the vendor clusters of the Aqara Spatial Multi-Sensor FP400 (vendor 0x115f, product 0x2009) to `custom-clusters` and the generated Python client, plus three fixes that were needed to make them usable end to end:

**Feature — Aqara FP400 clusters** (`packages/custom-clusters/src/clusters/aqara.ts`)
- `0x115ffc0a AmbientSensingConfiguration`: install settings (mode, side, height, sensitivity, …) and detection zones as a list of `{zoneId, zoneType, cells (16×20 cell bitmask), enabled}` structs with `SetZones` / `AppendZone` / `UpdateZone` / `RemoveZone` commands
- `0x115ffc0b RadarSensingUnion`: zone id on the zone child endpoints, human count, `MotionDetected` event
- `0x115ffc0c OccupantLocation`: activity state, `SubscribeLocationData` command and the `LocationInfo` event with per-target x/y positions

Names follow the trait names in the Aqara app's device model; values were verified against firmware 1.1.8.2. Note: the device only serves these clusters to fabrics whose vendor id it trusts (a fabric with the test vendor id 0xfff1 gets full access, 0x134b is answered with `UnsupportedAttribute`).

**Fix — Python generator for decorator-defined clusters** (`python_client/scripts/generate-python-clusters.ts`)
Decorator-defined clusters keep their type in `operationalBase` rather than a `type` string, so the generator fell back to the metatype: structs became `dict`, signed ints became `uint`, list entries and command responses were lost. The generator now resolves the operational base name, finds `listOf()` entries, emits cluster-local `Structs` and links `response()` classes as `<Request>Response`.
Side effect on existing output: `DraftElectricalMeasurementCluster.activePower` and `ThirdRealityMeteringCluster.instantaneousDemand` are `int32` in TS and were generated as `uint`; they are now `int`. Regenerating the whole tree is otherwise byte-identical.

**Fix — list conversion in WebSocket payloads** (`packages/ws-controller/src/server/Converters.ts`)
Lists of decorator-defined clusters carry `metabase.name === "list"` instead of `type === "list"`, so `isList()` did not recognise them and the base64 bytes inside `SetZones` reached the device as strings (`Expected bytes, got string`).

**Fix — `endpoint_added` ordering** (`packages/ws-controller/src/server/WebSocketControllerHandler.ts`)
When a node gains an endpoint at runtime (the FP400 creates one per zone), `endpoint_added` was sent synchronously while the `node_updated` snapshot was still scheduled via `setImmediate`. Home Assistant resolves the new endpoint from its node snapshot, raised a `KeyError` and dropped the connection. A pending `node_updated` for that node is now flushed before `endpoint_added`.

Commits are split per topic (generator fix, clusters + converter fix, endpoint ordering fix); happy to split them into separate PRs if preferred.

## Backing evidence

- Related issue: #
- [x] This fix/change is backed by a **complete log file** — attached here or in the linked issue (not just a few quoted lines). [matterjs-server-2026-09-19-evidence.log](https://github.com/user-attachments/files/32424475/matterjs-server-2026-09-19-evidence.log)

## Checklist

- [x] `npm test` passes (per package: custom-clusters, ws-controller, matter-server converters, python_client; the full root run stops at the pre-existing timezone-dependent `EnergyEvseUtilTest` failure on an untouched tree)


Apollon77:
@RAR thanks a lot, awesome PR. I think defining all the "enums" also as enums and reference them would make it even better? WDYT?

RAR:
> @RAR thanks a lot, awesome PR. I think defining all the "enums" also as enums and reference them would make it even better? WDYT?

I like it.  Let me poke around a bit and I'll give it a shot.  I don't think they'll make it over to the generated Python client due to how the generator works tho - it could likely be extended to emit real Python enums, however.   

The Aqara app is broken and won't connect over matter to the sensor at the moment plus there is a firmware update that won't install for the same reason - I'd like to fully verify everything prior to making this ready for review.

RAR:
Aqara unblocked the broken app non-sense so I was able to actually check my work from the app side.  Sir Claude worked through the enums as enums for me.   Currently trying to figure out why I can't write attributes, but the app can.

RAR:
Ok, good to go.   Did get a HA card kicking with the same layout setup that the app provides as well but I'm still working through getting it intuitive to use.

 
Screenshot 2026-09-21 at 1 43 48 PM

lboue:
@RAR Great to see the FP400 clusters land here, especially with the enum work and the writable region masks sorted out.

Quick question: has a companion PR been opened on the Home Assistant core repo yet to expose these new clusters (zones, occupancy, fall detection) as entities, or is that still on your todo list?

RAR:
> @RAR Great to see the FP400 clusters land here, especially with the enum work and the writable region masks sorted out.
> 
> Quick question: has a companion PR been opened on the Home Assistant core repo yet to expose these new clusters (zones, occupancy, fall detection) as entities, or is that still on your todo list?

I was waiting on this getting merged.  Would include whats currently included in the card repo: 

Screenshot 2026-09-21 at 1 44 58 PM

https://github.com/RAR/ha-aqara-fp400

